PR #2, opened 2018-01-15
Generate random walk parameters for knots of galaxy formation

EiffL:
This pull request add an interface to use the `knots` information provided by the new addon to the  protoDC2 catalog (see https://github.com/LSSTDESC/gcr-catalogs/pull/64). If the protoDC2_addon_knots catalog is used instead of the normal one, the flux of the  disk will be split between in two, and a portion of the flux will be attrtibuted to the random walk component.
The random walk component is stored in files using the `knots` spatial type and the same format as the sersic profile  (for simplicity) but the `sindex` parameter now corresponds to the number of knots.


jchiang87:
This looks good to me, but most of the changes are in code @danielsf wrote.   Scott:  Can you have a look when you get a chance?

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -22,12 +22,29 @@ class PhoSimDESCQA(PhoSimCatalogSersic2D, EBVmixin):
...
+
     # below are defined getter methods used to define CatSim value-added columns
     @cached
     def get_hasDisk(self):
         output = np.where(self.column_by_name('SEDs/diskLuminositiesStellar:SED_9395_583:rest')>0.0, 1.0, None)
         return output
 
+    @cached
+    def get_hasKnots(self):
+        return self.get_hasDisk()
```
danielsf: if you change this to

`return self.column_by_name('hasDisk')`

the catalog will cache the results of `get_hasDisk()` and return the cached values, rather than having to run it again (assuming it has already been run for its own sake).  This might not be necessary here, but is something you should know for developing on future CatSim code.

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -198,7 +203,29 @@ def _transform_catalog(self, gc):
         gc.add_quantity_modifier('sindex::disk', gc.get_quantity_modifier('sersic_disk'))
         gc.add_quantity_modifier('sindex::bulge', gc.get_quantity_modifier('sersic_bulge'))
 
-        return None
+        additional_postfix = ()
```
danielsf, line 221: Please add a few lines to the `_transform_catalog` docstring explaining what `additional_postfix` is.

python/desc/sims/GCRCatSimInterface/ProtoDC2DatabaseEmulator.py
```diff
@@ -228,6 +229,24 @@ def _transform_catalog(self, gc):
         gc.add_quantity_modifier('sindex::disk', gc.get_quantity_modifier('sersic_disk'))
         gc.add_quantity_modifier('sindex::bulge', gc.get_quantity_modifier('sersic_bulge'))
 
+        # Test for random walk specific addon
```
danielsf: Can you put this block of code into a `_transform_knots()` method defined in `DESCQAObject`, and then have both `DESCAObject._transform_catalog()` and `DESCQAObject_protoDC2._transform_catalog()` call `_transform_knots()`?  That way, we only have to worry about changing `_transform_knots()` if we end up having to develop on this code.

danielsf (approved):
A few minor change requests, but this looks good to merge.

danielsf:
I guess it looks good to merge except that it's going to need a `rebase`

EiffL:
Great :-) Thanks a lot Scott. I'm rebasing and making the modifications rightaway

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
EiffL: nice

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
EiffL: done

python/desc/sims/GCRCatSimInterface/ProtoDC2DatabaseEmulator.py
EiffL: very good suggestion, the code had already diverged between the two `_transform_catalog()` functions :-/ . Done

EiffL:
Let me quickly check that everything is still working with these last modifications

EiffL:
@danielsf Ok, everything still works, should be ok to merge

PR #3, opened 2018-01-23
Compound db object

danielsf:
As you may recall, in order to run the sprinkler in Twinkles, we had to create a structure that simultaneously queried the galaxy_disk, galaxy_bulge, and galaxy_agn tables on fatboy so that the sprinkler could self-consistently remove all of the components of the relevant galaxies.  This pull request extends that infrastructure to the GCRCatSimInterface.

The scripts `workspace/compound_db/make_compound_catalog.py` and `workspace/compound_db/verify_compound_catalog.py` show how this works.  The only major design changes is that `yaml_file_name`, `field_ra` and `field_dec` need to be defined in the `DESCQAObject` daughter classes at class definition time, rather than at instantiation.

`workspace/compound_db/verify_compound_catalog.py` verifies that this PR behaves the way it should by generating a PhoSim InstanceCatalog of bulges and a PhoSim InstanceCatalog of disks, and then generating both again with the CompoundCatalog framework and verifying that the contents of all three catalogs are consistent.  We can move this to a test directory, if we want to. It involves actually loading `proto-dc2_v2.1.2`, so I don't know if we want to treat it as an actual unit test or not (someday we should think about how to generate dummy GCR catalogs the same way we generate dummy SQL catalogs for unit testing purposes in `lsst_sims`).

Let me know what changes you think should be made before merging.

Also, Bryce Kalmbach should know about this, since he is actively working on extending the sprinkler now.  I emailed him.  He does not seem to be associated with this repo (at least, I cannot @ him or add him as reviewer).

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CompoundCatalogDBObjectClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CompoundCatalogDBObjectClasses.py
@@ -66,6 +66,14 @@ class CompoundDESCQAObject(DESCQAObject):
         super(CompoundDESCQAObject, self).__init__()
 
     def _make_column_map(self):
+        """
+        This dummy method is needed because the call to
+        super(CompoundDESCQAObject, self).__init__() in
+        this class's __init__ will try to call
+        self._make_column_map().  We have already done
+        the column map construciton in self.__init__()
+        and thus, do not want to do anything more.
+        """
         pass
 
     def _validate_input(self):
--- a/python/desc/sims/GCRCatSimInterface/CompoundCatalogDBObjectClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CompoundCatalogDBObjectClasses.py
@@ -8,53 +8,39 @@ __all__ = ["CompoundDESCQAObject"]
 
 class CompoundDESCQAObject(DESCQAObject):
     """
-    This is a class for taking several CatalogDBObject daughter classes that
-    query the same table of the same database for the same rows (but different
-    columns; note that the columns can be transformed by the CatalogDBObjects'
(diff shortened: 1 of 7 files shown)


jchiang87:
I added the Twinkles team to the repo.
@jbkalmbach

jbkalmbach:
Thanks Jim.

danielsf:
I should add that this code is dependent on this PR in `sims_catalogs`, which should be merged shortly

https://github.com/lsst/sims_catalogs/pull/18

jchiang87 (approved):
Looks ok to me, but I don't really know anything about these compound db objects to really assess this.  I confirmed that tests scripts ran ok and did a regression against the catalogs produced by master and everything checked out.  Adding @jbkalmbach as a reviewer.

jbkalmbach (approved):
Looks good. Once we have an AGN catalog to test I can implement this with the sprinkler.

jbkalmbach:
Is there any reason for us not to merge this now?

jchiang87:
@jbkalmbach Weren't you going to test this with an AGN catalog and the sprinkler first?  If you've already done that, then please go ahead and merge.

jbkalmbach:
Ok, I can wait and test on that before merging. If nobody else needs this that works for me.

danielsf:
Actually, I think the AGN sprinkler test is going to depend on a different PR I intend to issue tonight, which actually implements the unsprinkled AGN simulations.

Probably the right thing to do is merge this.  I will then rebase my new PR on top of it, and when *that* gets issued, Bryce can run his sprinkler test.

Sound good?

jbkalmbach:
Yeah, sounds good.

PR #12, opened 2018-01-25
Assign agn params

danielsf:
just move my changes to a branch of sims_GCRCatSimInterface


danielsf:
@jchiang87 @jbkalmbach This branch is now where we will work on this problem

jchiang87:
@jbkalmbach I'm trying to understand what additional work will need to be done on the scripts in bin.src folder to get all of this work to come together.   I see in your `twinkles_dc2_branch` branch that you've implemented `TwinklesCatalogWriter`.  Are you thinking that would replace `InstanceCatalogWriter` in `generateInstCat.py`?  There's a ton of replicated code as a result.  We can refactor later, but since I was looking into what would need to go into `generateInstCat.py` for this branch, we'd better make sure we are on the same page, so effort isn't wasted.

jbkalmbach:
I think once I can test with protoDC2 galaxies we can just have a setting where for visit that we want to use the sprinkler we use a different type of compoundDBObject. Are you planning on using the compoundInstanceCatalogs for the normal visits too? I don't see any reason not too.

jbkalmbach:
And this means that we can keep it all in `InstanceCatalogWriter` to further answer your question.

PR #14, opened 2018-01-25
Assign agn params rebase

danielsf:


danielsf:
@jbkalmbach @EiffL 

Could you please run whatever tests you have used to validate your work against this branch. I'm pretty sure I handled the rebase properly, but I've said that before and been wrong.

@EiffL The big thing I worry about with respect to your work is that I changed the code that detects whether or not `AlphaQAddOn` is available here

https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/assign_agn_params_rebase/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py#L18

and again in the `ProtoDC2DatabaseEmulator` module.

Once you both sign off, I will merge this.

yymao:
@EiffL at [`DatabaseEmulator.py#L282`](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/assign_agn_params_rebase/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py#L282),  instead of using `isinstance`, you can just set an attribute in your addon catalog and check for that attribute (using `getattr`), this way you don't have to import the base class `AlphaQAddOn`.

EiffL:
@danielsf @yymao Ok, I'm checking now, shouldnt take long, I also like Yao's idea

jbkalmbach:
Ok, seems like the AGN compound objects are working so I'm good with this. Once this is merged I will rebase my Twinkles branch on it and I can start getting twinkles to run with the proto_DC2 AGN.

EiffL:
@danielsf Ok, it works perfectly for me too, feel free to merge this.
I'll update the gcr so that you can check for the catalog type without resorting to `isinstance` and open a pull request on the master branch afterwards

PR #16, opened 2018-01-30
Twinkles dc2 branch

jbkalmbach:
Code to add sprinkler to DC2 Run 1.1


python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -159,6 +161,7 @@ def get_magNorm(self):
         raw_magnorm = self.column_by_name('magNorm_dc2')
         fitted_magnorm = self.column_by_name('magNorm_fitted')
         preliminary_output=np.where(np.isnan(raw_magnorm), fitted_magnorm, raw_magnorm)
+        preliminary_output = np.array(preliminary_output, dtype=np.float)
```
danielsf: You can achieve this by adding `.astype(float)` to the end of line 163.  Not sure if that will be faster, but I suspect that it will.

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -172,6 +172,8 @@ class DESCQAObject(object):
 
     objectTypeId = None
     verbose = False
+    tableid = 'galaxy'
```
danielsf: Why did you need to add these?

I *thought* I had implement the `CompoundDESCQAInstanceCatalog` class so that it evaluated the equality of `DESCQAObject`s on things that were relevant to `DESCQAObjects`.  Was something failing without these member variables?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -74,6 +80,14 @@ def __init__(self, opsimdb, descqa_catalog, dither=True,
                                host='fatboy.phys.washington.edu',
                                port=1433, driver='mssql+pymssql')
 
+        self.agn_db_name = os.path.join(os.environ['SCRATCH'], 'proto_dc2_agn',
```
danielsf: We will want this to be an argument passed in by the user so that Tom can redirect to wherever he puts that AGN parameters database for the final run.  `test_agn.db` is just a cartoon database I created for testing the AGN `DESCQAObject` during development.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -133,21 +147,49 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
...
-                          write_header=False)
-
-        disk_db = diskDESCQAObject(self.descqa_catalog)
-        disk_db.field_ra = self.protoDC2_ra
-        disk_db.field_dec = self.protoDC2_dec
-        cat = self.instcats.DESCQACat(disk_db, obs_metadata=obs_md,
-                                      cannot_be_null=['hasDisk'])
-        cat.write_catalog(os.path.join(out_dir, gal_name), chunk_size=100000,
-                          write_mode='a', write_header=False)
+        if self.sprinkler is False:
```
danielsf, line 151: We have not thought about how the sprinkler will interact with the knots catalog.

The knots code is meant to add more interesting structure to galaxy disks.  There is currently a non-zero probability that it will try to add knots to a galaxy that has been replaced by the sprinkler.  In that case, the knots will be added based on wrong assumptions about the size and SED of the galaxy.  Probably, the safest thing to do at this point is add code to prevent the knots catalogs from touching the sprinkled galaxies.  Maybe we can add a column to all `DESCQAObject`s named `sprinkled`.  This will be an int that defaults to 0, but gets set to 1 by the sprinkler.  We can then modify the knots catalog to ignore any galaxy with `sprinkled==1`.

Thoughts?

python/desc/sims/GCRCatSimInterface/TwinklesClasses.py
```diff
@@ -0,0 +1,202 @@
...
+twinkles_spec_map.subdir_map['(^specFile_)'] = twinkles_sn_sed_dir
+
+_twinkles_defs_file = os.path.join(os.environ['TWINKLES_DIR'], 'data', 'dc2_defs.csv')
+
+_agn_cache_file = os.path.join(os.environ['TWINKLES_DIR'], 'data', 'dc2_agn_cache.csv')
+_sne_cache_file = os.path.join(os.environ['TWINKLES_DIR'], 'data', 'dc2_sne_cache.csv')
+global_agn_params_db = os.path.join('/global/cscratch1/sd/brycek/',
+                                    'proto_dc2_agn', 'test_agn.db')
+
+assert os.path.exists(global_agn_params_db)
```
danielsf: Again: I think we should replace this with `if` block that raise exceptions explicitly telling the user that the relevant file doesn't exist

python/desc/sims/GCRCatSimInterface/TwinklesClasses.py
```diff
@@ -0,0 +1,202 @@
...
+assert os.path.exists(_agn_cache_file)
+assert os.path.exists(_sne_cache_file)
+
+__all__ = ["TwinklesCompoundInstanceCatalog_DC2",
+           "sprinklerCompound_DC2", "TwinklesCatalogSersic2D_DC2",
+           "TwinklesCatalogZPoint_DC2"]
+
+
+#class sprinklerCompound_DC2(sprinklerCompound, GalaxyCompoundDESCQAObject):
```
danielsf: Remove commented code

python/desc/sims/GCRCatSimInterface/TwinklesClasses.py
```diff
@@ -0,0 +1,202 @@
...
+    agn_params_db = global_agn_params_db
+    agn_objid = 'agn_descqa'
+
+    def _final_pass(self, results):
+        #From the original GalaxyTileCompoundObj final pass method
+        for name in results.dtype.fields:
+            if 'raJ2000' in name or 'decJ2000' in name:
+                results[name] = np.radians(results[name])
+
+        # the stored procedure on fatboy that queries the galaxies
```
danielsf: Go ahead and remove this block of comments, since this repo should never be used to get galaxies from fatboy

python/desc/sims/GCRCatSimInterface/TwinklesClasses.py
```diff
@@ -0,0 +1,202 @@
...
+                       defs_file=self.defs_file)
+        results = sp.sprinkle()
+
+        return results
+
+class TwinklesCatalogSersic2D_DC2(PhoSimDESCQA):
+
+    specFileMap = twinkles_spec_map
+
+#    column_outputs = ['prefix', 'uniqueId', 'raPhoSim', 'decPhoSim', 'phoSimMagNorm', 'sedFilepath',
```
danielsf: I would remove the commented out `column_outputs` in this and the class below

python/desc/sims/GCRCatSimInterface/TwinklesClasses.py
```diff
@@ -0,0 +1,202 @@
...
+                            use_it = False
+
+                        if use_it:
+                            compound_dbo = candidate(dbObjClassList)
+                            break
+
+                    if compound_dbo is None:
+                        compound_dbo = default_compound_dbo(dbObjClassList)
+
+                compound_dbo.mjd = self._obs_metadata.mjd.TAI
```
danielsf: Add a comment here pointing out why these lines are special

danielsf (changes requested):
This looks good to me.

The only reason I did not approve it is because I realized that we don't know what we are doing with respect to the knots objects that @EiffL added to the framework.  We should probably figure out how we are handling those before we mark this as 'Delivered'.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: @EiffL This is what I was referring to

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
jbkalmbach: I don't think I need those. I think I was trying to solve something in the compound Instance catalog and this was an attempted solution I didn't remove. I can't take those out now.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: Turns out, we won't be using knots in Run 1.1

@jbkalmbach: If you need someone to just approve this, you can ping me.

I will go back to minding my own business

EiffL: @danielsf  oh ok, I see, yes it makes a lot of sense to not add knots to the sprinkled galaxies. We can open an issue on this after the code gets merged if we don't want to delay the sprinkler at this point

danielsf (approved):
good to merge

PR #17, opened 2018-01-30
U/jchiang/command line script fixes

jchiang87:


python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -173,15 +173,14 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
             cat = self.instcats.DESCQACat_Agn(agn_db, obs_metadata=obs_md)
             cat.write_catalog(os.path.join(out_dir, gal_name), chunk_size=100000,
                               write_mode='a', write_header=False)
-
```
jchiang87, line 176: @danielsf The code from line 169 to this point works, but it also generates these messages to the screen:
```
Traceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 636, in _finalize_fairy
    fairy._reset(pool)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 774, in _reset
    self._reset_agent.rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1563, in rollback
    self._do_rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1601, in _do_rollback
    self.connection._rollback_impl()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 670, in _rollback_impl
    self._handle_dbapi_exception(e, None, None, None, None)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 188, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=exc_value)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 181, in reraise
    raise value.with_traceback(tb)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
sqlalchemy.exc.ProgrammingError: (sqlite3.ProgrammingError) Cannot operate on a closed database.
```
The code seems to run to completion correctly but it's not at all obvious there wasn't some failure.  For Run1.1, this isn't a pressing issue since we will always have the sprinkler enabled, but it would be good to fix whatever is causing this error message.

danielsf: I do so love sqlalchemy errors....

Is there a self-contained example script I can run to reproduce this?

jchiang87: meh,  let's put this off for now.   I'm not sure what it would take to distil this to smaller script....was hoping you'd seen this before.

danielsf: I associate the "cannot operate on a closed database" error with unit tests that don't clean up after themselves properly.  It's possible this is, as you say, innocuous.  Maybe sqlalchemy is trying to do something to the database after we're done with it and have manually closed/deleted it.

That is my off-the-cuff assessment.

jchiang87:
@jbkalmbach @danielsf It looks like the Sprinkler versions of the galaxy and agn classes being used in `InstanceCatalogWriter` are writing RA, Dec in units of radians instead of degrees to the instance catalogs.  Could one of you have a look please?  Might as well just put any fixes in this branch.  Thanks!

danielsf:
I have fixed the problem with RA, Dec.

There appears to be an additional problem with position angle.  I am showing values in excess of 1000, and it is supposed to be in degrees. This problem persists, even if the sprinkler is turned off.  I will continue to investigate.

danielsf:
And those last two commits fix the positionAngle problem.

jchiang87:
Thanks for the fixes @danielsf .  Here are plots showing the ~footprints of the object classes produced by the `generateInstCat.py` script with and without the sprinkler enabled for a 1deg fov centered on the dithered pointing position of visit 230 using `minion_1016_desc_dithered_v4.db`.  The higher count yellow regions are populated by the protoDC2 galaxies, while the lower count blue regions are just the stars.  The black points are my attempt to identify sprinkled SL objects by plotting entries with the same redshift as the previous entry and which have an agn SED specified.  I suspect the black points in the unsprinkled plot are accidently resampled AGN as consecutive entries.  I'm thinking it would be better to sample that population without replacement.

The region boundaries are from the Run 1.1 Specifications google doc.  The sprinkled region appears to be a bit smaller than intended, based on that doc.

sprinkled_v230_1deg-fov
unsprinkled_v230_1deg-fov

jchiang87:
I think we need to merge this before merging #19 .  Unless there are objections, I'll merge it into master at 12 Noon PT, today.

jbkalmbach:
Are people okay with the location of the sprinkled objects? I can add a new version of the caches if we want to get right up to the edge of the uDDF region. I left a little room on the edges on purpose just to absolutely make sure I stayed in the uDDF region. 

Also, Jim I think your RA boundaries for the uDDF in the plot are a little large (I think it should only go from 52.575 - 53.675 deg) making it a little more exaggerated in that direction.

jchiang87:
I got the boundary numbers from the Run1.1 Specifications document, so if the numbers there are wrong, that document should be corrected.  In any case, I'm ok with the current uDDF area for the sprinkled objects.

jbkalmbach:
Oh, no, you're right, I didn't see the table when I made the boundaries, I was just using 1.1 degree square sides so I guess there is that extra amount of space. Sorry about that. But if we're good I will just leave it alone for now.

katrinheitmann:
So should the document be corrected to be consistent?

jchiang87:
If the fix is easy, then please go ahead and make it, but I don't think it's crucial.  We have some time since we still need to look at #19.

If we don't change it, then we should update the Run1.1 spec with the as-implemented values.

jbkalmbach:
@katrinheitmann Maybe we could mention that the uDDF region is also a trapezium in the second paragraph of the design section.

katrinheitmann:
Please feel free to edit the document as you see fit. It's important 
that it describes exactly what we do :-)

jchiang87:
For the record, Bryce updated the cache files for the sprinkled lens objects.  Here is the updated plot of the sprinkled field:

sprinkled_v230_1deg-fov

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: I just accidentally triggered this message elsewhere.

It looks like it happens when the garbage collector tries to get rid of the `ObservationMetaDataGenerator`.  That is definitely new behavior.  I will fix it in the next `lsst_sims` release.

PR #19, opened 2018-01-31
Issue phosim catalog sn

rbiswas4:
Committed changes for writing out phosim catalogs of SN populations. 

## Inputs

There are data files in `data/` that represent SN in protoDC2 hosts in the Main Survey Region and the uDDF region separately. There are also three files that include SN that in hostless in the uDDF region and those in the  Main Survey at z < 1.0 and those in the Main Survey at z > 1.0 . These files are:
- MainSurveyHostedSNPositions_trimmed.csv 
- MainSurvey_hostlessSN_trimmed.csv       
- MainSurvey_hostlessSN_highz_trimmed.csv 
- uDDFHostedSNPositions_trimmed.csv
- uDDF_hostlessSN_trimmed.csv

They include the position and properties of the SN using which catsim can create images.

## The catalog classes
These are two new classes in `python/desc/sims/GCRCatSimInterface/CatalogClasses.py`
- `SNFileDBObject` : which uses catsim functionality to read from the csv files rather than fatboy which we did in Twinkles.
- `DC2PhosimCatalogSN` : Class to modify phosim catalogs for SN for the DC2 Run 1.1 .  A lot of this was mostly in Twinkles, but that seems hard to work with. So, there is some re-implementation.
## convenience functions 
There is a convenience function in `python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py`
- `snphosimcat` : uses the catalog classes to give a phosim instance catalog which can be written out.

## Demo : 
There is an example of how to use the phosim catalog in `
`Notebooks/demo_phosimcatalogs.ipynb`

## Storage:
The way the current method works is that it writes out a filename of the phosim spectrum to the instance catalog, but the spectra are written out to `spectra_files/Dynamic`. We can change this to point to the path we like, and we can also make sure that we don't have to put the entire path on the phosim instance catalog row.


rbiswas4:
@danielsf You know most of the code as you helped me getting through this. One thing I was a little worried about was whether I had to do any database unlinking. Currently, it does not look that way. Could you please take a look at that?

@TomGlanzman Can you please see if the storage mechanism we discussed is implemented to your liking. This does put spectra in a directory different from the usual one. Currently the phosim instance catalog only has a file name (because I recall we preferred shorter names) but that can be easily changed.

@jbkalmbach : Do you have strongly lensed SN at high redshifts and do you need to see them for long times? Maybe you did it differently, but I had some trouble with that. I had to change the `maxz` and Time visible parameters as in `python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py`

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
```diff
@@ -0,0 +1,170 @@
+"""
+A module for classes that read in the protoDC2 catalog and assign SN to hosts
+
+"""
+from __future__ import division, print_function, absolute_import
+__all__ =  ['DC2SN']
+import numpy as np
+import pandas as pd
+import snsims
```
jchiang87: This looks like a new dependency.  I guess it comes from [here](https://github.com/rbiswas4/SNsims)?  It should be added to [the prod software list](https://github.com/LSSTDESC/DC2_Repo/pull/120/files).

jchiang87: I also see another one:  https://github.com/rbiswas4/OpSimSummary.git

jchiang87: and also https://github.com/rbiswas4/AnalyzeSN

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
 from . import PhoSimDESCQA, PhoSimDESCQA_AGN
 from . import bulgeDESCQAObject_protoDC2 as bulgeDESCQAObject, \
     diskDESCQAObject_protoDC2 as diskDESCQAObject, \
-    knotsDESCQAObject_protoDC2 as knotsDESCQAObject, \
```
jbkalmbach: I think you accidentally removed the AGN and Twinkles imports.

python/desc/sims/GCRCatSimInterface/twinklesVariabilityMixins.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import absolute_import, division, print_function
```
jbkalmbach: Same comment as `twinklesCatalogDefs.py`

python/desc/sims/GCRCatSimInterface/twinklesCatalogDefs.py
```diff
@@ -0,0 +1,79 @@
+"""PhoSim Instance Catalog"""
```
jbkalmbach: Why can't you import this straight from desc.twinkles? I saw you mentioned this replication in your comment. Did I change something you needed in my Twinkles update and can we edit what you need in the Twinkles repo instead?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
+    # This means that the the spectra written by phosim will 
+    # go to `spectra_files/Dynamic/specFileSN_* 
+    # Note: you want DC2PhosimCatalogSN.sep to be part of this prefix
+    # string. 
+    # We can arrange for the phosim output to just read the string 
+    # without directories or something else
+
+    y.sn_sedfile_prefix = 'spectra_files/Dynamic/specFileSN_'
+    return y
+>>>>>>> added convenience function for doing SN populations
```
jchiang87: Please delete this line.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
jchiang87: Please restore the import of these symbols.

jbkalmbach (commented):
I decided when to include an object in the catalog differently I think so I didn't have a `maxTimeSNVisible`. I just created the SED and looked if there was > 0 flux in the imsimband.

jchiang87 (changes requested):
I'm still testing the execution of the code with changes merged from #17, but I wanted to get these comments in before the files change.

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,15 +4,91 @@
...
-__all__ = ["PhoSimDESCQA", "PhoSimDESCQA_AGN"]
+from lsst.sims.catalogs.db import fileDBObject
+
+__all__ = ["PhoSimDESCQA", "PhoSimDESCQA_AGN", "SNFileDBObject",
+           "DC2PhosimCatalogSN"]
 
 #########################################################################
 # define a class to write the PhoSim catalog; defining necessary defaults
 
+class SNFileDBObject(fileDBObject):
```
danielsf: The `SNFileDbObject` class should actually go in `DatabaseEmulator.py`.  That is where we are keeping the classes that behave like `CatalogDBObject`s.

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,15 +4,91 @@
...
+               ('Tt0', 't0_in'),
+               ('Tx0', 'x0_in'),
+               ('Tx1', 'x1_in'),
+               ('Tc', 'c_in'),
+               ('id', 'snid_in'),
+               ('Tredshift', 'z_in'),
+               ('redshift', 'z_in'),
+               ('Tgaltileid', 'galtileid')
+              ]
+class DC2PhosimCatalogSN(PhoSimCatalogSN):
```
danielsf: There should be space between class definitions (but this problem will, of course, go away if you move `SNFileDBObject` to `DatabaseEmulator.py`)

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,15 +4,91 @@
...
+    """
+    Modification of the PhoSimCatalogSN mixin to provide shorter sedFileNames
+    by leaving out the parts of the directory name. Also fix name changes from
+    gamma to shear.
+    """
+    def get_sedFilepath(self):
```
danielsf, line 25: Do we really need to map `TsedFilepath` into `sedFilepath`, given that the catalog is just going to output `shorterFileNames`, anyway?  Shouldn't `get_shorterFileNames` just call `self.column_by_name('TsedFilepath')`?

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,15 +4,91 @@
...
+
+    # column_outputs = PhoSimCatalogSN.column_outputs
+    # column_outputs[PhoSimCatalogSN.column_outputs.index('sedFilepath')] = \
+    #    'shorterFileNames'
+    column_outputs = ['prefix', 'uniqueId', 'raPhoSim', 'decPhoSim',
+                      'phoSimMagNorm', 'shorterFileNames', 'redshift',
+                      'shear1', 'shear2', 'kappa', 'raOffset', 'decOffset',
+                      'spatialmodel', 'internalExtinctionModel',
+                      'galacticExtinctionModel', 'galacticAv', 'galacticRv']
+    cannot_be_null = ['x0', 't0', 'z', 'shorterFileNames']
```
danielsf: Can you put a blank line between `column_outputs` and `cannot_be_null` (for readability)?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: It's probably more subtle than that.  I'm going to *guess* that these were added after Rahul's branch was created.  Rebasing to `master` (after Jim merges #17 ) will probably fix this....

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
-
-__all__ = ['InstanceCatalogWriter', 'make_instcat_header', 'get_obs_md']
+    knotsDESCQAObject_protoDC2 as knotsDESCQAObject
+from . import DC2PhosimCatalogSN, SNFileDBObject
+
+__all__ = ['InstanceCatalogWriter', 'make_instcat_header', 'get_obs_md',
+           'snphosimcat']
+
+
+SNDTYPESR1p1 = np.dtype([('snid_in', int),
```
danielsf: Can you add a comment explaining what this is for?

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,15 +4,91 @@
...
+    columns = [('raJ2000', 'snra_in*PI()/180.'),
+               ('decJ2000', 'sndec_in*PI()/180.'),
+               ('Tt0', 't0_in'),
+               ('Tx0', 'x0_in'),
+               ('Tx1', 'x1_in'),
+               ('Tc', 'c_in'),
+               ('id', 'snid_in'),
+               ('Tredshift', 'z_in'),
+               ('redshift', 'z_in'),
+               ('Tgaltileid', 'galtileid')
```
danielsf: You can probably drop this `('Tgaltileid', 'galtileid')` mapping, since it isn't used.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
+                         ('snra_in', float),
+                         ('sndec_in', float)])
+
+def snphosimcat(fname, tableName, obs_metadata, objectIDtype, idColKey='snid_in',
+              delimiter=',', dtype=SNDTYPESR1p1):
+    """convenience function for writing out phosim instance catalogs for
+    different SN populations in DC2 Run 1.1 that have been serialized to
+    csv files.
+
+    Parameters: string
```
danielsf: should `fname` appear in this line of the docstring?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
+	absolute path to csv file for SN population.
+    tableName : string 
+	table name describing the population upto user choice.
```
danielsf: typo "upto"

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
+    obs_metadata: instance of `lsst.sims.utils.ObservationMetaData`
+	observation metadata describing the observation
+    idColKey : string, defaults to values for Run1.1
+        describes the input parameters to the database as interpreted from the
+        csv file.
+    delimiter : string, defaults to ','
+        delimiter used in the csv file
+    dtype : instance of `numpy.dtype`
+	tuples describing the variables and types in the csv files.
+    """
```
danielsf: Can you add a "returns" section to the docstring?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -21,13 +21,66 @@
...
+    
+    x = SNFileDBObject(fname, runtable=tableName,
```
danielsf: I'd feel more comfortable using more explanatory variable names than `x` and `y`.  Maybe `dbobj` and `cat`?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: This looks like some cruft left over from a rebase error....

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class DC2SN(object):
+    """ Class to associate SN in the SN params table with hosts
+    
+    Parameters
+    ----------
+    galsdf : pandas dataFrame
+        provides galaxy table
+    snpops : instance of SNPopulation
```
danielsf: is `SNPopulation` defined in `snsims`?

jchiang87:
@rbiswas4 I'm trying to understand how the SN sed files will be generated for Run1.1 as well as for later production runs that will use this code.  It looks like the `demo_phosimcatalogs.ipynb` is meant to provide some guidance, but I think we either need a script in `bin.src` that does this in a self explanatory way (similar to `create_agn_db.py`) or we need real documentation, e.g., a fully annotated notebook or an .rst document.  If I'm missing something, please let me know where I should be looking.

Also, operationally, this code doesn't seem to be hooked up to `generateInstCat.py` via `InstanceCatalogWriter`.  When I run `generateInstCat.py` after merging from #17, the resulting instance catalogs look identical to what I would get from #17 as-is.

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
```diff
@@ -0,0 +1,170 @@
+"""
```
danielsf: Adding this class (and the jupyter notebooks that use it) adds a dependence on `snsims`, which brings with it a dependence on `esheldon/fitsio` and `OpSimSummary`.  Do we want `sims_GCRCatSimInterface` to depend on those packages, or, given that the infrastructure being added here just reads in the supernova populations from a csv file, should another repository be created that contains the code to actually generate those csv files, while the repository just takes it on faith that the csv files being supplied are correct?

danielsf (changes requested):
There are some organization changes that need to be made.

I also just want to open up the question of whether or not we want this package (`sims_GCRCatSimInterface`) to depend on Rahul's`snsims`package, which in turn depends on `esheldon/fitsio` and Rahul's `OpSimSummary` package.  These packages are all just used to generate the csv files containing the supernova populations for DC2.  Should that code go into a different repository?

If we do want to keep that code here, we need to update the README of this repository so that people know how to go about installing those other packages.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: possibly ... .There was a conflict, and I mistakenly believed that this was a line that was not on master. Will Fix.

python/desc/sims/GCRCatSimInterface/twinklesCatalogDefs.py
rbiswas4: Actually, I made things indpendent. Had meant to delete those. Will delete them.

python/desc/sims/GCRCatSimInterface/twinklesVariabilityMixins.py
rbiswas4: same as before.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: will fix. Sorry.

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
rbiswas4: I actually want to keep this somewhere else. This will not be required for production.

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
rbiswas4: Scott: That is a mistake. I don't want to keep this in this repository. It is not going to be used in production

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
rbiswas4: Don't know. Not planning to change something that is working at this point. Happy to look at it later if you want to keep this as an issue.

danielsf: OK.  We can look at this as we refactor things for Run 2.0

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -47,10 +49,12 @@ def snphosimcat(fname, tableName, obs_metadata, objectIDtype, idColKey='snid_in'
...
 
-    Parameters: string
-	absolute path to csv file for SN population.
+    Parameters:
+    -----------
+    fname : string
+        absolute path to csv file for SN population.
     tableName : string 
-	table name describing the population upto user choice.
+        table name describing the population to be descided by user choice.
```
danielsf: Typo "descided"

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -117,6 +182,21 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
         knots_name = 'knots_cat_%d.txt' % obsHistID
         #agn_name = 'agn_cat_%d.txt' % obshistid
 
+	# SN Data
+        snDataDir = '../../../../data'
```
danielsf: Can you replace this with
```
os.path.join(getPackageDir('sims_GCRCatSimInterface'), 'data')
```
where `getPackageDir` can be imported from `lsst.utils`?

That way, the code will be able to find the `csv` files, no matter where it is run from (assuming everything has been set up properly via EUPS)

danielsf:
I was able to get the InstanceCatalogWriter to run using this branch (I will explain how to run this test below).  It runs, however:

- The way we are constructing PhoSim InstanceCatalogs, we are separating each class of object (stars, galaxies, AGN) into separate catalog files and then including them in the main PhoSim catalog using PhoSim's `includeobj` command.  The supernova catalogs do not get included this way.  The catalog names need to be added to the `object_catalogs` kwarg of `make_instcat_header` in `InstanceCatalogWriter.write_catalog()`.

- The supernova catalog generator produces five distinct catalogs:
        -MainSurveyHostedSNPositions_cat_230.txt 
        -MainSurvey_hostlessSN_cat_230.txt 
        -MainSurvey_hostlessSN_highz_cat_230.txt
        -uDDFHostedSNPositions_cat_230.txt
        -uDDF_hostlessSN_cat_230.txt
  What are the differences between these?  Which should actually be included in the InstanceCatalog?

To run this test, you will need the script `workspace/instcat_writer/test_instance_catalog_writer.py` which is on `master`.  It will run through the whole InstanceCatalog generation process and write the results to `workspace/instcat_writer/sprinkled/` (which it will create, if needed).  The paths to the OpSim database and the agn_parameter database are hard-coded to copies that live in my `SCRATCH` space.  I have given these `755` permissions, so if you just replace `os.environ['SCRATCH']` with `'/global/cscratch1/sd/danielsf'` it should work.  You also need to increase `fov` in `write_catalog()` to at least `0.5` to get non-empty supernova catalogs.  You will also need to create the SED directory before running this script.

rbiswas4:
- will try adding names of the SN catalogs. 

> The supernova catalog generator produces five distinct catalogs:
-MainSurveyHostedSNPositions_cat_230.txt
-MainSurvey_hostlessSN_cat_230.txt
-MainSurvey_hostlessSN_highz_cat_230.txt
-uDDFHostedSNPositions_cat_230.txt
-uDDF_hostlessSN_cat_230.txt
What are the differences between these? Which should actually be included in the InstanceCatalog?

According to what I have said (SNIa at the natural rate  + enhanced in uDDF region but still not exceeding the natural rate of SN (ie if you consider core collapse) : all of them. However, I realize that a reasonable chunk comes from MainSurvey_hostlessSN_highz_cat_230.txt which is likely not very useful. So, if there is a crunch this could be left off without really affecting people. Hence I separated the highz and lowz. Aside from the optional nature of `-MainSurvey_hostlessSN_highz_cat_230.txt` (z > 1.0), everything should be there. The reason for having hostless SN is to try to have SN that do not have a galaxy background, on the same set of exposures (actually hopefully ccds)  to compare during validation. The point  is if things are bad at hosts that has a different source than being bad for hostles ones. The reason for having them in different catalogs is to have different objectId, and identify the origin easily.

- I will try the test after fixing the paths.

rbiswas4:
@danielsf I made the changes but get permission denied to try and run your script.

rbiswas@cori04:~/src/DC2/sims_GCRCatSimInterface/workspace/instcat_writer> python test_instance_catalog_writer.py
OPSIM /global/projecta/projectdirs/lsst/groups/SSim/DC2/minion_1016_desc_dithered_v4.db
/global/cscratch1/sd/danielsf/proto_dc2_agn/test_agn.db
Traceback (most recent call last):
  File "test_instance_catalog_writer.py", line 16, in <module>
    assert os.path.exists(agn_db_name)
AssertionError

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -26,8 +27,73 @@
...
+    cat.phoSimHeaderMap = DefaultPhoSimHeaderMap
+    cat.writeSedFile = True
+
+    # This means that the the spectra written by phosim will 
+    # go to `spectra_files/Dynamic/specFileSN_* 
+    # Note: you want DC2PhosimCatalogSN.sep to be part of this prefix
+    # string. 
+    # We can arrange for the phosim output to just read the string 
+    # without directories or something else
+
```
jchiang87: So it appears that the SED files will be written to a subdirectory `spectra_files/Dynamic` relative to the current working directory, correct?  If that works ok for Tom's phosim pipeline, I guess that is fine.  Could we add a line here like
```
    os.makedirs('./spectra_files/Dynamic', exist_ok=True)
```
?  Since the path is not selectable by the user, we might as well make sure the directory exists before writing to it.  This would be helpful since this part of the sprinkler is only executed after the non-sprinkled objects have been written to disk.  That may take a couple of tens of minutes depending on the fov value, and it would be frustrating to have the code crash after all that time simply because that directory didn't already exist.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -192,6 +276,15 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
 
             gal_cat.write_catalog(os.path.join(out_dir, gal_name), chunk_size=100000,
                                   write_header=False)
+        
+        # SN instance catalogs
+        for i, snpop in enumerate(snpopcsvs):
+            phosimcatalog = snphosimcat(snpop, tableName=names[i],
+                                        obs_metadata=obs_md, objectIDtype=i+42)
+
+            snOutFile = name +'_cat_{}.txt'.format(obsHistID)
```
jchiang87: looks like this should be
```
           snOutFile = names[i] +'_cat_{}.txt'.format(obsHistID) 
```

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -117,10 +183,28 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
...
+
+        snpopcsvs = list(os.path.join(snDataDir, n) for n in 
+                        [sncsv_hostless_uDDF,
+                         sncsv_hostless_pDC2,
+                         sncsv_hostless_pDC2hz,
+                         sncsv_hosted_uDDF,
+                         sncsv_hosted_pDC2])
+
+        names = list(snpop.split('/')[-1].split('.')[0].strip('_trimmed')
+                         for snpop in snpopcsvs)
```
jchiang87, line 210: We should build the `object_catalogs` more carefully.  Let's insert the following here:
```
        object_catalogs = [star_name, gal_name] + \
                          ['{}_cat_{}.txt'.fomat(x, obsHistID)
                           for x in names[:5]]
```

jchiang87 (changes requested):

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -117,10 +183,28 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
...
         make_instcat_header(self.star_db, obs_md,
                             os.path.join(out_dir, cat_name),
                             imsim_catalog=self.imsim_catalog,
-                            object_catalogs=(star_name, gal_name))
+                            object_catalogs=(star_name, gal_name, names[0],
```
jchiang87: lines 206-207 should be instead
```
                           object_catalogs=object_catalogs)
```

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -192,6 +276,15 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
...
+        # SN instance catalogs
+        for i, snpop in enumerate(snpopcsvs):
+            phosimcatalog = snphosimcat(snpop, tableName=names[i],
+                                        obs_metadata=obs_md, objectIDtype=i+42)
+
+            snOutFile = name +'_cat_{}.txt'.format(obsHistID)  
+            phosimcatalog.write_catalog(os.path.join(out_dir, snOutFile),
+                                        chunk_size=10000, write_header=False)
 
         if self.imsim_catalog:
```
jchiang87, line 299: line 296 below should be
```
         for orig_name in object_names:
```
so that the SN files get gzipped.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: Done

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: Done.

jchiang87:
With the changes I've proposed I am able to get a set of phosim instance catalog files that look right.  The following command line works at NERSC:
```
time generateInstCat.py --db /global/projecta/projectdirs/lsst/groups/SSim/DC2/minion_1016_desc_dithered_v4.db \
--agn_db_name /global/cscratch1/sd/jchiang8/DC2_prs/agn_db_mbh_7.0_m_i_30.0.sqlite \
--descqa_catalog protoDC2 \
--out_dir sprinkled_sn_sl_0.5 \
--ids 230 \
--fov 0.5 \
--protoDC2_ra 55.064 \
--protoDC2_dec -29.783 \
--enable_sprinkler
```
The code will need some fixing to produce an imsim catalog and will need some other fixes, but once my suggested changes are addressed, I think it will be ready for Run1.1p.

danielsf:
Stupid thing: do we need to add `twinkles` to the `ups/sims_GCRCatSimInterface.table` file?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: This is better, but I should probably drop the `5` in case the size of the list is changed. It is the complete list.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: Done

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
rbiswas4: I don't see the `object_names` variable somewhere.

jchiang87: Sorry, should be the `object_catalogs` list made above.

rbiswas4: Done

jchiang87 (approved):

jchiang87:
With @rbiswas4 's most recent change, I was able to run the code to completion using the command line I posted last night.  There are a bunch of formatting fixes I'd like to have done and refactoring to make everything work more coherently (and to fix the imsim catalog generation), but unless those issues prevent us from convincing ourselves that the current code will work for Run1.1p, I propose we merge.
@danielsf Can you give the code a look?   We may need to make another set of changes in case the SED file handling won't work for @TomGlanzman 's phosim pipeline, but I'd prefer to do that as an isolated change on a separate branch/pr.

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -4,16 +4,66 @@
...
+                fname = 'None'
+            split_names.append(fname)
+        return np.array(split_names)
+
+    # column_outputs = PhoSimCatalogSN.column_outputs
+    # column_outputs[PhoSimCatalogSN.column_outputs.index('sedFilepath')] = \
+    #    'shorterFileNames'
+    column_outputs = ['prefix', 'uniqueId', 'raPhoSim', 'decPhoSim',
+                      'phoSimMagNorm', 'shorterFileNames', 'redshift',
+                      'shear1', 'shear2', 'kappa', 'raOffset', 'decOffset',
```
danielsf, line 53: I just want to explicitly state that we are not going to be applying weak lensing parameters to these supernovae, since there are no weak lensing parameters in `SNFileDBObject`.  `shear1`, `shear2`, and `kappa` will always return their default values of 0.0 in the case of supernovae.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -26,8 +27,74 @@
...
+    """convenience function for writing out phosim instance catalogs for
+    different SN populations in DC2 Run 1.1 that have been serialized to
+    csv files.
+
+    Parameters:
+    -----------
+    fname : string
+        absolute path to csv file for SN population.
+    tableName : string 
+        table name describing the population to be decided by user choice.
```
danielsf, line 58: `objectIDtype` is not mentioned in the docstring

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -195,6 +282,15 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
 
             gal_cat.write_catalog(os.path.join(out_dir, gal_name), chunk_size=100000,
                                   write_header=False)
+        
+        # SN instance catalogs
+        for i, snpop in enumerate(snpopcsvs):
+            phosimcatalog = snphosimcat(snpop, tableName=names[i],
+                                        obs_metadata=obs_md, objectIDtype=i+42)
+
+            snOutFile = names[i] +'_cat_{}.txt'.format(obsHistID)
```
danielsf, line 294: In the future, I would like to consider creating the supernova catalog names in their own list that can be appended to `[star_name, gal_name]`.  We can then use that list of supernova catalog names here.  That will reduce the likelihood of lines 291 and 209 drifting out of sync.

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
danielsf: @rbiswas4 I thought you were going to remove this file (and, presumably, the notebooks that depend on it)?

danielsf (approved):
This is good to merge.  I would prefer that `simCat_DC2sndb.py` and the notebooks that depend on it be removed before we do so, but that is not a hard requirement.

I also just want to make sure: it does not appear that the weak lensing shear map data is going to be applied to supernovae.  Is this intentional/are we aware of this?

jbkalmbach:
Should we have a single variable to set somewhere that will tell the SN and sprinkled SN SEDs to go to the same place?

jbkalmbach:
Or should we just wait to hear back from Tom and handle this all then?

danielsf:
@jbkalmbach Probably, but I say merge and wait for feedback from Tom.

This branch has already been rebased too many times ;)

jchiang87:
> Should we have a single variable to set somewhere that will tell the SN and sprinkled SN SEDs to go to the same place?

@jbkalmbach I'm not quite understanding this.  I thought only the code that Rahul added in this branch is writing SEDs (to `./spectra_files/Dynamic`).  Are there other SED files being written somewhere?

jbkalmbach:
Yes, if sprinkled SN exist the SEDs get written to the folder set [here](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/TwinklesClasses.py#L11)

jchiang87:
> Yes, if sprinkled SEDs exist they get written to the folder set here

And the writing of those files is handled in the Twinkles classes?  I assume that must be relative to the cwd, right?  So the result will be a directory tree that looks like
```
spectra_files/
|-- Dynamic
|   |--specFileSN_1086_59580.1396_r.dat
|   |-- specFileSN_12039_59580.1396_r.dat
|   |-- specFileSN_12296_59580.1396_r.dat
...
|
`-- <Sprinkled SN files>
```
Is that right?

jbkalmbach:
Yeah, that's what it should look like.

jchiang87:
ok, good. then I say we merge and make a new branch in case we need to adjust the SED writing.

python/desc/sims/GCRCatSimInterface/CatalogClasses.py
rbiswas4: - As I said I am happy to remove them. If there is a suggestion of where to track such products, I will follow that. Or else keep these in a separate repository for my own tracking.

- About weak lensing: Yes, I know about this. I have been thinking that I would go with unlensed SN in Run 1.1 and make sure that the SN land where I put them wrt the protoDC2 galaxies. My guess is that if I provide these values in the csv files, then it would be trivial to turn them on in Run 1.2, at least with the shear values provided in protoDC2. At some level, I don't want too many unknowns to validate simultaneously.

TomGlanzman:
I've a question about the SED files location.  Referring to the argument descriptions in Jim's interface script, sims_GCRCatSimInterface/bin.src/generateInstCat.py, can someone describe in detail the structure of the "out_dir" created?   I am interested in the names and locations of the main instanceCatalog file, included files, the SED directory naming and content.  If a test has been run at NERSC for which I could examine the output, that would be great.

jchiang87:
Here is the resulting directory structure from that command:
```
[DC2_prs] tree sprinkled_sn_sl_0.5
sprinkled_sn_sl_0.5
|-- MainSurveyHostedSNPositions_cat_230.txt.gz
|-- MainSurvey_hostlessSN_cat_230.txt.gz
|-- MainSurvey_hostlessSN_highz_cat_230.txt.gz
|-- bright_stars_230.txt
|-- gal_cat_230.txt.gz
|-- knots_cat_230.txt
|-- phosim_cat_230.txt
|-- star_cat_230.txt.gz
|-- uDDFHostedSNPositions_cat_230.txt.gz
`-- uDDF_hostlessSN_cat_230.txt.gz

0 directories, 10 files
```
The instance catalogs are all in `/global/cscratch1/sd/jchiang8/DC2_prs/sprinkled_sn_sl_0.5` and the SEDs are in `/global/cscratch1/sd/jchiang8/DC2_prs/spectra_files`

TomGlanzman:
Thank you @jchiang87 and @rbiswas4 .  In addition, Rahul provided me with a full example output and I observed an issue with the SED specification within the catalog files.  We are slacking to hammer this out.

jchiang87:
ok, I will merge in ~30 minutes, unless I hear otherwise.

rbiswas4:
Scott pointed out a hole in documentation that I will fix. And, there is the question of whether you have a preference for storing the files that should not be there.

jchiang87:
Thanks @rbiswas4.  ETA for merging is still 11:30 am PT.

python/desc/sims/GCRCatSimInterface/simCat_DC2sndb.py
rbiswas4: Yes. I am going to run `git rm` . If you have a suggestion of where to keep this that is good. Otherwise I will put them in my repository.

danielsf: It seems like these scripts are still in a state where only you can run them (i.e. they depend on some of your other repositories).  In that case, I would say: keep them in your own personal repository for now.  We can work on making these scripts publicly usable at another time.

rbiswas4:
@jchiang87 ready to merge

jchiang87:
ok, I'm doing a final git pull, will run the tests and the command line and if all looks ok, will merge....should be ~10mins.

TomGlanzman:
Rahul and I have settled on a solution that will work for PhoSim (and the DC2 workflow).  While he is working on the code and running a test, I will describe.

The "out_dir" specification will define a directory which will include (as it does now) all of the catalog files.  In addition, it will contain a new subdirectory, "Dynamic" which will contain all of the SED files.

References to these dynamic SED files from the catalog files will look something like this:

object 3401774 53.4327981 -27.2598065 23.5489169 Dynamic/specFileSN_3322_59580.1396_r.dat 0.42006242 0 0 0 0 0 point CCM CCM 0.0275142564 3.1

In order to convince phoSim to find these files, I will create a symlink mirror of the production SED directory in $SCRATCH, e.g., $SCRATCH/DC2/\<visitID\>/SEDs.  Then, will add a single sym link named "Dynamic" from there to the out_dir/Dynamic directory.  Finally, phoSim will be run specifying the $SCRATCH area for all of its SEDs.  This scheme will allow an arbitrary number of simultaneous and distinct phosim instances to run in parallel.

jchiang87:
@TomGlanzman Can you make a new github issue for that discussion please? Or wait for Rahul's PR and put that text there? Thanks.

jchiang87:
So after doing the git pull, I tried to rerun that command line and the code is failing with sqlalchemy errors.  I don't see how the last two commits could possibly cause an issue, but I think we need to get this working before merging.

jchiang87:
```
ceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: unable to open database file

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/global/cscratch1/sd/jchiang8/DC2_prs/sims_GCRCatSimInterface/bin/generateInstCat.py", line 51, in <module>
    instcat_writer.write_catalog(obsHistID, out_dir=args.out_dir, fov=args.fov)
  File "/global/cscratch1/sd/jchiang8/DC2_prs/sims_GCRCatSimInterface/python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py", line 287, in write_catalog
    write_header=False)
  File "/global/cscratch1/sd/jchiang8/DC2_prs/sims_GCRCatSimInterface/python/desc/sims/GCRCatSimInterface/TwinklesClasses.py", line 173, in write_catalog
    write_mode=write_mode)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sims_catalogs/2.4.2.sims/python/lsst/sims/catalogs/definitions/CompoundInstanceCatalog.py", line 355, in _write_compound
    for chunk in master_results:
  File "/global/cscratch1/sd/jchiang8/DC2_prs/sims_GCRCatSimInterface/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py", line 140, in __next__
    return self._descqa_obj._postprocess_results(chunk)
  File "/global/cscratch1/sd/jchiang8/DC2_prs/sims_GCRCatSimInterface/python/desc/sims/GCRCatSimInterface/ProtoDC2DatabaseEmulator.py", line 295, in _postprocess_results
    chunk_size=1000000)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sims_catalogs/2.4.2.sims/python/lsst/sims/catalogs/db/dbConnection.py", line 470, in get_arbitrary_chunk_iterator
    return self.get_chunk_iterator(query, chunk_size = chunk_size, dtype = dtype)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sims_catalogs/2.4.2.sims/python/lsst/sims/catalogs/db/dbConnection.py", line 484, in get_chunk_iterator
    return ChunkIterator(self, query, chunk_size, arbitrarySQL = True)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sims_catalogs/2.4.2.sims/python/lsst/sims/catalogs/db/dbConnection.py", line 78, in __init__
    self.exec_query = dbobj.connection.session.execute(query)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/orm/scoping.py", line 150, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/orm/session.py", line 1023, in execute
    bind, close_with_result=True).execute(clause, params or {})
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 914, in execute
    return meth(self, multiparams, params)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/sql/elements.py", line 323, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1010, in _execute_clauseelement
    compiled_sql, distilled_params
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1146, in _execute_context
    context)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 188, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=exc_value)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 181, in reraise
    raise value.with_traceback(tb)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file [SQL: 'SELECT galaxy_id, magNorm, varParamStr FROM agn_params WHERE galaxy_id BETWEEN 10936 AND 11311777 ORDER BY galaxy_id']
Exception during reset or similar
Traceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
sqlite3.ProgrammingError: Cannot operate on a closed database.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 636, in _finalize_fairy
    fairy._reset(pool)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 774, in _reset
    self._reset_agent.rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1563, in rollback
    self._do_rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1601, in _do_rollback
    self.connection._rollback_impl()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 670, in _rollback_impl
    self._handle_dbapi_exception(e, None, None, None, None)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 188, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=exc_value)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 181, in reraise
    raise value.with_traceback(tb)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
sqlalchemy.exc.ProgrammingError: <unprintable ProgrammingError object>
Exception during reset or similar
Traceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
sqlite3.ProgrammingError: Cannot operate on a closed database.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 636, in _finalize_fairy
    fairy._reset(pool)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 774, in _reset
    self._reset_agent.rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1563, in rollback
    self._do_rollback()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1601, in _do_rollback
    self.connection._rollback_impl()
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 670, in _rollback_impl
    self._handle_dbapi_exception(e, None, None, None, None)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 188, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=exc_value)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/util/compat.py", line 181, in reraise
    raise value.with_traceback(tb)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 668, in _rollback_impl
    self.engine.dialect.do_rollback(self.connection)
  File "/global/common/software/lsst/cori-haswell-gcc/stack/w.2017.46_py3_gcc6/stack/miniconda3-4.3.21-10a4fa6/Linux64/sqlalchemy/1.0.8.lsst3+4/lib/python/SQLAlchemy-1.0.8-py3.6-linux-x86_64.egg/sqlalchemy/engine/default.py", line 420, in do_rollback
    dbapi_connection.rollback()
sqlalchemy.exc.ProgrammingError: <unprintable ProgrammingError object>

real	3m27.158s
user	1m13.383s
sys	0m18.401s
```

danielsf:
@jchiang87 ....grumble... I suspect this on the lsst_sims side.

I will try to fix it.

jchiang87:
I think I see the issue actually....some files have appeared in my directory with Rahul's uid and gid for the agn params db.

danielsf:
Oh good...  There is a lunch talk here I would like to attend.

I will check back in at 1:30 PST to see if a rush fix is needed

jchiang87:
The issue was temporary files created by sqlite that were hanging around under Rahul's userid.  (see https://sqlite.org/tempfiles.html for possible mitigations in the future).  The test ran fine after cleaning up those files.  So proceeding with the merge.

PR #22, opened 2018-02-02
Changes to modify sed filepath as described in issue:

rbiswas4:
- Added a parameter to the factory method `snphosimcat` to pass in the
  script output directory to set directory of spectra relative to that.
- Changed the `sn_sedfile_prefix` of the created catalogs to combine the
  input `sedRootDir` into the directory of spectra
- Pass the output directory `out_dir` to the `snphosimcat` factory
  functions from `InstanceCatalogWriter`
- formatting changes
	modified:   InstanceCatalogWriter.py
- Changed the separator variable in  `getsthorterFileNames` method of `D2PhosimCatalogSN` mixin
	modified:   CatalogClasses.py
This is the SN part. We also need this for lensed SN.  @jbkalmbach

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -35,7 +35,7 @@ class DC2PhosimCatalogSN(PhoSimCatalogSN):
         a phosim catalog.
         """
         fnames = self.column_by_name('sedFilepath')
-        sep = 'specFileSN_'
+        sep = 'Dynamic/specFileSN_'
         split_names = []
         for fname in fnames:
             if 'None' not in fname:
--- a/python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
+++ b/python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
@@ -189,7 +189,7 @@ class InstanceCatalogWriter(object):
 
         obs_md = get_obs_md(self.obs_gen, obsHistID, fov, dither=self.dither)
         # Add directory
-        twinkles_spec_map.subdir_map['(^specFile_)'] = out_dir
+        twinkles_spec_map.subdir_map['(^specFile_)'] = str(os.path.join(out_dir, 'Dynamic'))
 
         cat_name = 'phosim_cat_%d.txt' % obsHistID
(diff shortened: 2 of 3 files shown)


TomGlanzman (approved):
The result of @rbiswas4 testing showed an acceptable output directory structure containing both catalog files and SED files.  Further, the SED file references contained within the catalog files can be integrated with phoSim and its associated DC2 workflow.

jchiang87:
For lensed SN, it looks like one can just add the following to the imports in `InstanceCatalogWriter.py`:
```
from . import twinkles_spec_map
```
and then in `InstanceCatalogWriter.write_catalog(...)` insert after line 187
```
        twinkles_spec_map.subdir_map['(^specFile_)'] = out_dir
```
Since `twinkles_spec_map` is held as a reference in the various classes in `TwinklesClasses.py`, this should propagate that directory information to all of the classes that need it.

jchiang87 (changes requested):
I think the changes I described in my comment should do the trick for lensed SNe.

rbiswas4:
@jchiang87  @jbkalmbach  `twinkles_spec_map` is not imported, so I imported it from the file directly, and tried testing. But I am not sure I can find the lensed SN rows in phosim catalogs yet

rbiswas4:
The lensed SN are in uDDF only .. duh. So, I should rerun the script with center at uDDF. Right @jbkalmbach ?

jbkalmbach:
Yeah, that's the way to do it.

jchiang87:
@rbiswas4 Did you identify a visit where the lensed SN SEDs were written out?  I'd like to try it out to so I can make a plot showing the locations of the various sprinkled sources in the Run1.1 region for Phil's talk.

jbkalmbach:
I'm working on this right now. For some reason the sprinkled SN are not appearing.

jbkalmbach:
Ok, I've figured out the issue and will be pushing changes shortly.

rbiswas4:
@jchiang87 I think this is fixed with Twinkles (work in 5721869) in the branch  `correct_dc2_caches` and this branch

jchiang87 (approved):

jchiang87:
@rbiswas4 I was able to run the visit Bryce recommended using the most recent commits, so if you guys think it's ready go ahead and merge.  I'd like to do some final checks on everything tonight before tagging.

PR #27, opened 2018-02-13
add script to plot distributions of instance catalog columns

jchiang87:
As a sanity check on the instance catalog files created by #26, I wrote a script to plot distributions of quantities like RA, Dec, magnorm, galaxy shape parameters, and WL distortion parameters.  Here is the output running on our standard sprinkled visit `obsHistID=197356` for a 1 deg radius fov:
phosim_cat_197356
There are still some oddities with the galaxy shape parameters.  4601 galaxies have minor axis values `<= 0`, and one galaxy has a major/minor axis ratio `> 8000`.  Here is its entry:
```
object 18313778273 53.585852 -27.8084637 28.5890625 galaxySED/Inst.50E09.02Z.spec.gz 0.830407977 0.00905066356 0.00321041653 -0.00566315651 0 0 sersic2d 0.113193986 1.37278762e-05 135.216739 4 CCM 0.205300367 4.70105981 CCM 0.0259158766 3.1
```

### DIFF
--- a/bin.src/plot_instcat_dists.py
+++ b/bin.src/plot_instcat_dists.py
@@ -12,15 +12,15 @@ import matplotlib.pyplot as plt
 
 class HistArray(object):
     "Class to manage subplotting of histograms."
-    def __init__(self, title='', figsize=(9, 12), shape=(5, 2)):
+    def __init__(self, title='', figsize=(12, 12), shape=(4, 3)):
         """
         Parameters
         ----------
         title: str ['']
             Overall title of the figure.
-        figsize: tuple [(9, 12)]
+        figsize: tuple [(12, 12)]
             Figure dimensions in x, y inches
-        shape: tuple [(5, 2)]
+        shape: tuple [(4, 3)]
             Number of subplots in the y and x dimensions, respectively.
         """
         plt.rcParams['figure.figsize'] = figsize
@@ -58,7 +58,7 @@ class HistArray(object):
         plt.yscale(yscale)
         plt.legend(loc=1, fontsize=6)
 
-def plot_instcat_dists(phosim_file, figsize=(9, 12)):
+def plot_instcat_dists(phosim_file, figsize=(12, 12)):
     """
     Create a multipanel plot of histograms of various columns in the
     phosim instance catalog.
@@ -83,6 +83,8 @@ def plot_instcat_dists(phosim_file, figsize=(9, 12)):
     magnorm = defaultdict(list)
     major_axis = defaultdict(list)
     minor_axis = defaultdict(list)
+    pa = defaultdict(list)
+    sersic_index = defaultdict(list)
     num_zero_major = 0
     num_zero_minor = 0
     axis_ratio = defaultdict(list)
@@ -112,6 +114,8 @@ def plot_instcat_dists(phosim_file, figsize=(9, 12)):
                                                 minor_axis[item][-1])
                         if axis_ratio[item][-1] > 1000:
                             print(line.strip())
+                    pa[item].append(float(tokens[15]))
+                    sersic_index[item].append(float(tokens[16]))
                     gamma1[item].append(float(tokens[7]))
                     gamma2[item].append(float(tokens[8]))
                     kappa[item].append(float(tokens[9]))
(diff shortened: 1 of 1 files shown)


katrinheitmann:
@jchiang87 Hi Jim, do you think it would make sense to add this to DESCQA -- similar to the readiness test we have for the extra-galactic catalog? Or is that overkill? Eve and Dan should take a look at the major/minor axis ratio issue, but I can't seem to add them here via @. Maybe they are not part of this channel? Thanks!

jchiang87:
I do think it would be good to add these tests to DESCQA.  However, I still think we need a standalone script for developers to use while they are working on the instance catalog generation code.

I added the Simulation Validation Framework and CosmoSim teams to this repo so that we can ping @evevkovacs.

yymao:
I'm guessing that the galaxies that have weird minor/major size ratios could be those galaxies that do not have a bulge in the extragalactic catalog. I am not sure how CatSim deals with those cases, but it may result in strange numbers. For galaxies that have finite bulge sizes, I think they all have sensible minor/major size ratios. 

On a totally different note, @jchiang87 I don't think those GitHub teams are up-to-date. For one, there's only 6 members in CosmoSim and I am not one of them... I've talked to @drphilmarshall to see how we can improve this situation, because right now, in most cases, using teams is just confusing.

evevkovacs:
@danielsf should comment on how CatSim deals with cases where the bulge component is zero. It is perfectly valid to have such a case. I just checked the output of the readiness test for protoDC2 and the minimum bulge size is ~1e-3 in the extra-galactic catalog (ie no negative values). The minimum size (bulge + disk) is slightly larger. See https://portal.nersc.gov/project/lsst/descqa/v2/?run=2018-01-30_4&test=readiness_protoDC2&right=2018-01-30_4/readiness_protoDC2/protoDC2/SUMMARY.html  I think any negative values are being introduced in the CatSim processing. On a technical note, wouldn't it be better to check minor/major rather than major/minor since the latter could have extreme values for edge-on galaxies?

evevkovacs:
@yymao  @jchiang87  @danielsf I think we should add a check for minor/major to the readiness test for protoDC2 so it can be compared with the CatSim output.

jchiang87:
@yymao The github teams can be managed via the DESC member database automatically.  See
https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=199820385

yymao:
@jchiang87 yes they can. But most WGs/teams do not have that set up. 

@evevkovacs we can add a quantities like `aspect_ratio_disk` and `aspect_ratio_bulge` if we want to include them in the readiness test. Though we already have checked the aspect ratios are consistent with ellipticities. And we have checked ellipticity distribution...

jchiang87:
@evevkovacs In the instance catalog I examined in the figure, I think all of the minor and major axis values filtered for `<=0` were actually zero and not negative.  I previous iterations of the instance catalog script using protoDC2, I saw negative axis values.  I can switch the ratio to minor/major, but maybe a scatter plot of major vs minor would be useful.  This major/minor > 8000 case seems extreme no matter how it's viewed and plotting minor/major distributions would hide its presence in the plot.

jchiang87:
@yymao There is a mechanism for managing those teams so that they are up-to-date. We don't need to bother Phil in order to use this mechanism.

yymao:
@jchiang87 maybe we should move this discussion somewhere else rather than in this issue... My point is just that, right now, DESC GitHub teams are not being used effectively and routinely result in confusion. I've already talked to Phil so the bothering part is done anyway...

danielsf:
Currently, the block of code to detect whether a galaxy has a disk or bulge component is here

https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L92

It actually detects on the flux in one of the tophat filters, rather than morphology parameters.  A few weeks ago, there was discussion of changing this behavior so that bulge/disk existence was tested on morphology.  That has not been implemented, yet.

danielsf:
The object with that large major/minor axis ratio is the spheroid of the galaxy with galaxy_id 17884549

I have checked that the large axis ratio does exist in the raw protoDC2 catalog.

dkorytov:
The bulge axis ratios are drawn from a Gaussian centered at 0.8 with a width of 0.2 and clipped to be in the range 0->1 (to roughly follow galaxy zoo's distribution in fig 4, https://arxiv.org/pdf/1306.3264.pdf). So having a 1/8000 axis ratio is possible. I'm not sure where the zeros are coming in for the minor sizes. As @evevkovacs said, the readiness tests show that they are all above zero. 

The plot below has bulge axis ratios in green and disk axis ratios in blue. 
fig3

This is the distribution of axis ratios of bulge and disks:
fig2

jchiang87:
If there are no objections, I will merge this into master tomorrow morning (Feb 21, 9am PT).

PR #28, opened 2018-02-16
change proto-dc2_v2.1.2 to protoDC2

### LINKED ISSUE #18: use 'protoDC2' instead of 'proto-dc2_v2.1.2'
When getting the catalog from `GCRCatalogs`, it would be better to use `'protoDC2'` instead of `'proto-dc2_v2.1.2'`, so that when `'proto-dc2_v2.1.2'`become obsolete in a future version of `GCRCatalogs`, this interface will also receive the same update. 

Note that, when using a frozen version of `GCRCatalogs`, `'protoDC2'` will *always* point to the same catalog, even if the upstream has updates (it'll only print out a warning message). So there is no need to worry about catalog version not being controlled when using `'protoDC2'`, as it only updates with the `GCRCatalogs` package. 

Here's the current occurrence of `'proto-dc2_v2.1.2'` in master:
```
workspace/compound_db/verify_compound_catalog.py
13:    yaml_file_name = 'proto-dc2_v2.1.2'

workspace/compound_db/make_compound_catalog.py
11:    yaml_file_name = 'proto-dc2_v2.1.2'

tests/test_full_instanceCatalogs.py
40:        db = diskDESCQAObject_protoDC2(yaml_file_name='proto-dc2_v2.1.2')
66:        db = diskDESCQAObject_protoDC2(yaml_file_name='proto-dc2_v2.1.2')

bin.src/create_agn_db.py
86:                        default='proto-dc2_v2.1.2',
88:                        "Default = 'proto-dc2_v2.1.2'")

workspace/agn_params/test_phosim_agn_catalog.py
13:    db = agnDESCQAObject_protoDC2(yaml_file_name='proto-dc2_v2.1.2')

workspace/agn_params/test_agn_catalog.py
22:    db = agnDESCQAObject_protoDC2(yaml_file_name='proto-dc2_v2.1.2')

workspace/agn_params/test_agn_compound.py
18:    yaml_file_name = 'proto-dc2_v2.1.2'
```

yymao:
This PR fixes #18 by replacing all occurrence of `proto-dc2_v2.1.2` to `protoDC2`. See the discussion in #18 for details.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/ProtoDC2DatabaseEmulator.py
+++ b/python/desc/sims/GCRCatSimInterface/ProtoDC2DatabaseEmulator.py
@@ -151,7 +151,7 @@ class DESCQAObject_protoDC2(DESCQAObject):
     _cat_cache_suffix = '_rotated'
     tableid = 'galaxy'
     database = 'LSSTCATSIM'
-    yaml_file_name = 'proto-dc2_v2.1.2'
+    yaml_file_name = 'protoDC2'
 
     def _rotate_to_correct_field(self, ra_rad, dec_rad):
         """
--- a/tests/test_full_instanceCatalogs.py
+++ b/tests/test_full_instanceCatalogs.py
@@ -37,7 +37,7 @@ class BulgePhoSimCatalogTestCase(unittest.TestCase):
         ObservationMetaData, using protoDC2 (to make sure we don't
         break the whole interface)
         """
-        db = diskDESCQAObject_protoDC2(yaml_file_name='proto-dc2_v2.1.2')
+        db = diskDESCQAObject_protoDC2(yaml_file_name='protoDC2')
         db.field_ra = self.field_ra
         db.field_dec = self.field_dec
         obs = ObservationMetaData(pointingRA=self.field_ra+0.2,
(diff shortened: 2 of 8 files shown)


jchiang87 (approved):
These changes are ok with me, but for the test code, since asserts are made assuming specific numbers, I would have thought it better to have the actual catalog name rather than the alias since a remapping of the alias may cause those tests to fail when in fact there is no error in the code.

yymao:
@jchiang87 Thanks. 

For the test issue you mentioned, here's my view: If we have a new catalog that is very different from `protoDC2`, we probably should have a new name for the catalog (i.e., not calling it `protoDC2`) so that wouldn't break the code here. Updates to `protoDC2` should be mostly bug fixes or minor adjustments, and these minor updates should not break the tests. If they do, it would probably be a good idea to know anyways. Does that make sense?

jchiang87:
That's all fine with me.  These are tests written by @danielsf, so I defer to his judgement.

danielsf (approved):
The tests in question are all "does it run?" tests that don't really depend on the detailed contents of the catalog.

You would have to try very hard to break them by updated protoDC2.

Go ahead and merge.

PR #29, opened 2018-02-17
use stellar mass to check galaxy component (fix #10)

### LINKED ISSUE #10: more robustly check for disk-only or bulge-only galaxies
Right now we use the SED to check if the galaxy has only disk or only bulge (see [`python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L27`](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L27)). 

I think the suggestion from @dkorytov is to use the disk/bulge size.

This should be a straightforward change too and I am also happy to take on it. @danielsf, comments or thoughts?

yymao:
A simple PR that fixes #10, i.e., to use stellar masses instead of SEDs to check if a galaxy component exists.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -28,7 +28,7 @@ class DC2PhosimCatalogSN(PhoSimCatalogSN):
     def get_shorterFileNames(self):
         """
         Method to truncate filenames for transient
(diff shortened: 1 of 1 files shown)


danielsf (approved):
Looks fine to merge

yymao:
I think this is also ready to be merged. Thanks!

PR #30, opened 2018-02-17
get sed labels from GCRCatalogs (fix #8)

### LINKED ISSUE #8: update SED labels to conform to GCR schema
While this can be as simple as just replace the content in the `data/dc2_seds.txt`, I feel like we can use this chance to refactor the first part of `data/get_sed_mags.py`: instead of getting the SED labels from a text file, we can directly obtain them from GCR directly.

@danielsf any comments or suggestions? I'll be happy to take a stab on this.

yymao:
As discussed in #8, this PR removes the static file `data/dc2_seds.txt` and use GCRCatalogs' native method to obtain the list of SED labels. This PR fixes #8.


danielsf (changes requested):
If we are going to do this, we should probably make `get_sed_mags.py` write a header which records what tophat filters were actually used in constructing the grid.  The `sed_from_galacticus_mags` method should then verify that the grid it is reading in uses the filters corresponding to the magnitudes it is being passed.  This would prevent a user from accidentally using the wrong magnitude grid to fit SEDs (thinking ahead to the day when we are using this infrastructure for more than just one extragalactic catalog).

yymao:
@danielsf is the runtime of `get_sed_mags.py` long? I am thinking we can even just merge `get_sed_mags.py` into `SedFitter.py` and avoid the intermediate file `CatSimMagGrid.txt`. Is that feasible?

danielsf:
`get_sed_mags.py` takes 44 seconds to run, so: no.

Do we have to worry about many separate catalog-generation processes trying to read in the galaxy SED files at once?  I honestly don't know.  There are just under a thousand galaxy SEDs in the library.

jchiang87:
I think in the current phosim pipeline, there will be one job running the instance catalog generation code per visit; and those jobs can, in general, run simultaneously.  @TomGlanzman can comment on how many he expects in practice to run at the same time, but if the overhead is only 44 seconds (compared to ~20-30mins currently for the Run1.1p catalogs on cori-haswell), I think it may make sense to avoid generating `CatSimGridMag.txt` as an intermediate file.

TomGlanzman:
Yes, there is a single instanceCatalog generation instance per visit.  I have been told that the UW catalog server can handle ~50 (or maybe a few more) simultaneous queries.  To be a good citizen, my default plan has been to limit the number of concurrent instanceCatalog generation steps to ~30.

Is there a way the UW queries could be reported, along with time-stamps in the log?  That would better help me understand if/when bottlenecks might occur.

yymao:
I have merged `get_sed_mags.py` into `SedFitter.py`. The code has changed significantly so I need to test it to make sure it works as intended. Will report back soon.

danielsf:
@TomGlanzman Are you asking for some log information from the UW server, or something on the CatSim side?

python/desc/sims/GCRCatSimInterface/SedFitter.py
```diff
@@ -1,74 +1,104 @@
...
+    for q in gc.list_all_quantities():
+        m = _gcr_sed_re.match(q)
+        if m:
+            wav0, width, tag = m.groups()
+            bp_params_raw[tag].add((int(wav0), int(width)))
+    assert bp_params_raw['disk'] == bp_params_raw['bulge'], 'SEDs for disk and bulge do not match'
+    assert bp_params_raw['disk'], 'No SED found'
+
+    bp_params_sorted = sorted(bp_params_raw['disk'], key=lambda p: p[0])
+    wav_min = bp_params_sorted[0][0] * 0.1
```
danielsf: Add a comment pointing out that this 0.1 is there because CatSim thinks in nm and the GCR thinks in Angstroms.  I just spend an embarrassingly long time trying to figure out why my own code requires that ;)

danielsf (approved):
Looks good to me

TomGlanzman:
@danielsf I was thinking something on the NERSC side which basically emitted a date/time stamp prior to and just after a communication with the UW server.  The reason being to have some raw data should any questions of throughput, or contention arise.  And it would be useful to have some timing information in any case for future planning.

yymao:
OK, I have tested the code and it is ready to be reviewed/merged.

danielsf (approved):
looks good

danielsf:
I'll merge this at noon Pacific Time, if no one objects

Thanks, Yao

PR #32, opened 2018-02-20
use dust parameters in GCR schema (fix #9)

### LINKED ISSUE #9: Use dust parameters Av and Rv from GCR
Per the discussion in #5 and https://github.com/LSSTDESC/gcr-catalogs/issues/70, the dust parameters Av and Rv will be included in the input catalog and become available in GCR, so this interface can (and should) directly use them.

I think this should be straightforward, basically revising (removing?) the function `get_internalDustParams` in [`python/desc/sims/GCRCatSimInterface/CatalogClasses.py`](https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/master/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L37).

@danielsf any thoughts? I think I can also help with this one.

yymao:
Another simple PR to fix #9. We now use the dust parameters as defined in the GCR schema. 

Current GCR schema only has one-component dust parameters. But this PR is future compatible in the sense that the code would use two-component dust parameters when exist; and if not, fall back to one-component dust parameters. 

Thanks to @danielsf for some technical help.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -28,7 +28,7 @@ class DC2PhosimCatalogSN(PhoSimCatalogSN):
     def get_shorterFileNames(self):
         """
         Method to truncate filenames for transient
-        spectra written out by phosim. 
+        spectra written out by phosim.
 
         .. note: the variable sep needs to be in
         `self.sn_sedfile_prefix` before writing out
@@ -39,7 +39,7 @@ class DC2PhosimCatalogSN(PhoSimCatalogSN):
         split_names = []
         for fname in fnames:
             if 'None' not in fname:
-                fname = sep + fname.split(sep)[-1] 
+                fname = sep + fname.split(sep)[-1]
             else:
                 fname = 'None'
             split_names.append(fname)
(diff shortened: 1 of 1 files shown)


python/desc/sims/GCRCatSimInterface/CatalogClasses.py
```diff
@@ -28,7 +28,7 @@ def get_sedFilepath(self):
     def get_shorterFileNames(self):
         """
         Method to truncate filenames for transient
-        spectra written out by phosim. 
+        spectra written out by phosim.
```
danielsf, line 31: Didn't you change these whitespace errors on another PR I approved this morning?

....I guess we'll find out when we merge.

danielsf (approved):
This all looks good.

@jchiang87: my understanding is that protoDC2 has not yet implemented different dust parameters for different components of the same galaxy.  If we merge this now, we will thus be changing the reddening in PhoSim/ImSim until such time as the new protoDC2 catalog comes out.  Am I correct that we are safe to do this because the version of `sims_GCRCatSimInterface` that will be run for Run 1.1p has already been frozen?

jchiang87:
The current code for Run1.1p is frozen at v0.1.1 of this repo, but unfortunately, we didn't get things working completely for imSim catalog generation.  So if we merge this branch, then get the imsim catalogs working, there will be this difference in the reddening of Run1.1i vis-a-vis Run1.1p.   Since there are or will be other differences between the two sims (e.g., we may have random walk galaxies in Run1.1i), I don't see this as a real concern.  Others' mileage may vary.

jchiang87 (approved):
lgtm

yymao:
In fact, after https://github.com/LSSTDESC/gcr-catalogs/pull/83 is merged in GCRCatalogs, the new code here will then behave exactly the same as the original...

danielsf:
Okay.   I will merge this at noon Pacific Time, then, if no one objects

PR #33, opened 2018-03-07
U/danielsf/ref cat

danielsf:
this PR adds the script to generate reference catalogs for DC2

### DIFF
--- a/bin.src/generate_dc2_reference_catalog.py
+++ b/bin.src/generate_dc2_reference_catalog.py
@@ -326,7 +326,7 @@ if __name__ == "__main__":
                                              'projectdirs',
                                              'lsst', 'groups',
                                              'SSim', 'DC2',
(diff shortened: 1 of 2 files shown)


jchiang87:
@danielsf Is there a way of saving the extragalactic catalog name (the actual one, not just an alias) as metadata in the reference catalog and potentially in its downstream derivations?  As we've seen, having that info more readily available would make provenance tracking much easier.

danielsf:
Yes.  I will add that.

danielsf:
Note, I am going to close this PR and move to `u/danielsf/refCat2`.  There was some non-trivial rebasing needed to get this branch built on top of master.

evevkovacs:
@jchiang87 @danielsf The extra-galactic catalog has a version number included in the metadata, which could/should be transferred to the instance catalog, if I am understanding correctly what you are asking for.

PR #34, opened 2018-03-29
U/danielsf/debug/dust columns

danielsf:
This change is needed to prevent a crash because numpy arrays returned directly by the Generic Catalog Reader are read-only.

We should investigate whether the code to deal with nonsense Av and Rv values is still necessary, now that we are past protoDC2-v2.1.2

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -1,5 +1,6 @@
 import os
 import numpy as np
+import copy
 from .SedFitter import sed_from_galacticus_mags
 from lsst.utils import getPackageDir
 from lsst.sims.catUtils.exampleCatalogDefinitions import PhoSimCatalogSersic2D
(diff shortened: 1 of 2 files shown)


jchiang87 (approved):
This looks good.  Was there a change in gcr-catalogs that necessitated this?  I've been bumping into this problem while refactoring `InstanceCatalogWriter.py` for eventual imsim use, but somehow wasn't getting bitten by it using current master.

yymao:
@jchiang87 GCRCatalogs does not guarantee the numpy array is writable though it usually is. However, for some file format and for some quantity modifiers, the returned array would be read-only.

danielsf:
More specifically: we used to get Av and Rv by querying the GCR for extincted and unextincted magnitudes, subtracting them and storing the difference in a numpy array that we created, which is thus writeable.  Yao recently updated the GCR to get Av and Rv directly from the extra-galactic catalog, yielding a numpy array that may or may not be writeable.

I do not know why you were not running into this problem on `master`.  Yao's change to the GCR was several weeks ago.

Issue #40, opened 2018-04-02
Remove dependency on CatSimMagGrid.txt


jchiang87:
#30 removed the `get_sed_mags.py` script that generates `CatSimMagGrid.txt` but the `CatalogClasses.get_fittedSedAndNorm` still depends on that text file.

jchiang87:
closed by #41

PR #41, opened 2018-04-03
U/danielsf/remove/cat sim mag grid

danielsf:
This issue

https://github.com/LSSTDESC/sims_GCRCatSimInterface/issues/40

notes that we did not completely remove the dependence on the pre-computed grid of CatSim magnitudes.  The InstanceCatalog classes still expected it to be there.  This PR updates the InstanceCatalog classes so that they automatically detect which bandpasses to use for SED fitting.


yymao (commented):
@danielsf I am a bit confused about this PR. Much of the added code in `get_fittedSedAndNorm` seems to be doing exactly the same thing as the function `_get_sed_mags_and_cosmology`.

danielsf:
The InstanceCatalog class needs to know the names of the luminosity columns it is asking GCR for, which is why I recreated the `re` code from `_get_sed_mags_and_cosmology`.  `_get_sed_mags_and_cosmology` itself loads the column names and calculates the magnitudes of every SED in the library in those bandpasses.  I did not want to do that twice.  If you would prefer, we can split out the part of `_get_sed_mags_and_cosmology` that finds the column names into a separate method and call that both from the InstanceCatalog class and from `_get_sed_mags_and_cosmology`.

yymao:
Unless I missed something, it seems to me that Lines 155 to 215 of `CatalogClasses.py` is doing the same thing as Lines 29 to 43 of `SedFitter.py`. Maybe we can change `_get_sed_mags_and_cosmology` a bit so that it can take either a string or a loaded gcr instance?

danielsf:
It is doing the same thing.  The reason is that, by the time the code gets here

https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/u/danielsf/remove/CatSimMagGrid/python/desc/sims/GCRCatSimInterface/CatalogClasses.py#L234

it needs to know the names of the catalog columns that contain the fluxes we are interested in.

If you want, I can write a separate method that just returns the names of those columns.  I can then have CatalogClasses and `_get_sed_mags_and_cosmology` use that new method.  Would that alleviate your concerns?

yymao:
I guess we can let `_get_sed_mags_and_cosmology` return the column names as well? 

But then looking at this again I wonder if that's the best way. As far as I can tell, the function `sed_from_galacticus_mags` is only called by `get_fittedSedAndNorm`, right? So why doesn't `sed_from_galacticus_mags` just read in the columns directly (i.e., not taking `mag_array` as an argument)?

danielsf:
I wanted `sed_from_galacticus_mags` to be a utility function that could be used generally to fit SEDs to a set of galacticus mags.

I am working on my proposed refactor.  Expect a few new commits in about an hour.

danielsf:
This is the refactor I was talking about.

danielsf:
For what it's worth: when we transition to CosmoDC2 and the entire catalog is too big to load in memory, we are going to need a way to tell `sed_from_galacticus_mags` which objects to fit at any given time.  In the current design, CatSim will handle querying the DC2 galaxies one chunk at a time so that `sed_from_galacticus_mags` won't have to worry about that.

yymao:
@danielsf OK. There's several things that I think we can still improve to make the code more clear but I am not sure if it's worth the time to implement them. 
- `sed_filter_names_from_catalog` should check if `wav_min` and `wav_width` are the same for disks and bulges and just returns for arrays: `wav_min`, `wav_width`, `disk_name`, and `bulge_name` (since it already verifies the filters are the same). 
- `_get_sed_mags_and_cosmology` can then be rename to `fit_sed_mags`. It can take in only `wav_min` and `wav_width` as input, and returns `sed_names`, `sed_mag_list`, and `sed_mag_norm`. The idea is that this function simply translates the top hat filters to the sed library without doing other things and so it does not need to know anything about the catalog etc.)
- Finally, `sed_from_galacticus_mags` will be called as `sed_from_galacticus_mags(galacticus_mags, redshift, wav_min, wav_width, H0, Om0)`. When this function is called for the first time, it called `fit_sed_mags(wav_min, wav_width)` and cache `sed_names`, `sed_mag_list`, and `sed_mag_norm`. For subsequent calls it just checks if both `wav_min` and `wav_width` stay the same.

danielsf:
`sed_filter_names_from_catalog` doesn't actually verify that the bulges and disks have the same tophat filters.  That is done in `_get_sed_mags_and_cosmology`.  I think I would like to leave `sed_filter_names_from_catalog` as it is, in case we are ever in a situation where the bulges and disks do have different tophat filters.

I agree that `_get_sed_mags_and_cosmology` has become unwieldy.  I will implement your suggestion for that method and `sed_from_galacticus_mags` and we can see if we like it.

danielsf:
Okay.  I think this is what you asked for.

yymao (approved):
Looks good. Thanks!

danielsf:
Thanks, Yao.

@jchiang87 I will merge this around 6pm PDT, unless I hear otherwise

jchiang87:
Thank to you both for handling this.   @danielsf I did a quick regression test and everything works as before, so please go ahead and merge when you see fit.

PR #43, opened 2018-04-19
sed label names change in GCRCatalogs v0.7.4

yymao:
Starting from GCRCatalogs v0.7.4, the SED label names that CatSim needs have been changed. See https://github.com/LSSTDESC/gcr-catalogs/issues/104 https://github.com/LSSTDESC/gcr-catalogs/issues/105 for details.

This PR makes the corresponding changes in this interface.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/SedFitter.py
+++ b/python/desc/sims/GCRCatSimInterface/SedFitter.py
@@ -32,8 +32,8 @@ def sed_filter_names_from_catalog(catalog):
     All outputs will be returned in order of increasing wav_min
     """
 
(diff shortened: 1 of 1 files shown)


danielsf (approved):
Looks good.

Do we need to be careful about interfering with people actually running `sims_GCRCatSimInterface` at the moment, or will the addition of the `no_host_extinction` columns to `protoDC2_v3` automatically be detected?

yymao:
Since the production only uses tagged versions, these changes won't take effect until the next time we update the list of tagged versions. So I think we are ok.

yymao:
@jchiang87 am I right that we can merge this without worrying affecting the production? If so, can you merge this? Thanks.

PR #45, opened 2018-04-24
replace Av == nan galaxies with random Av values

danielsf:


jchiang87 (approved):
I verified that this fixes the issue for my test case using proto-dc2_v3.0.  Note that I used the current master of gcr-catalogs in order for this code to work, i.e., v0.7.2 would not work for me.

PR #46, opened 2018-04-27
U/danielsf/ref cat2

danielsf:
This PR will add a script to generate a reference catalog for use with DC2.

### DIFF
--- a/bin.src/generate_dc2_reference_catalog.py
+++ b/bin.src/generate_dc2_reference_catalog.py
@@ -326,7 +326,7 @@ if __name__ == "__main__":
                                              'projectdirs',
                                              'lsst', 'groups',
                                              'SSim', 'DC2',
-                                              'agn_db_mbh_7.0_m_i_30.0.sqlite'))
+                                             'agn_db_mbh_7.0_m_i_30.0.sqlite'))
 
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


danielsf:
@evevkovacs @jchiang87 I assume the name of the full `yaml` file will be more useful than the version number, in case we ever start generating catalogs against buzzard or some other simulation.  Is there a reason I shouldn't use the name of the `yaml` file?

yymao:
@danielsf I think we should use the name of the yaml file. The reason we want to use the alias is to avoid having to update multiple places when the catalog is updated. But you can still keep track of what catalog the alias pointed to though.

evevkovacs:
@yymao @danielsf The yaml file name would work, but I guess I have a preference for using the catalog name and version as provided by the catalog metadata. This is more formal, and unambiguous. I can imagine yaml files coming and going, different abbreviations/names being used etc, whereas what's in the metadata is permanent and unchanging. Yao will probably have an opinion.

danielsf:
Actually, I just found the `GCRCatalogs.get_catalog_config()` method.  I'll just output the contents of that dict.  That is probably the most detailed option.

danielsf:
(this may, in fact, be what Eve was advocating, anyway)

yymao:
@danielsf yes, I think keeping track of the config content is the best solution here.

PR #49, opened 2018-05-21
U/danielsf/sprinkled/truth

danielsf:
This isn't actually a mature pull request, so much as a place where we can organize our thoughts on how to get truth information out of the sprinkler.

I have added an `is_sprinkled` column to the InstanceCatalogs.  There is a PR into the Twinkles repo which will alter the sprinkler to set `is_sprinkled=1` for any objects changed or added by the sprinkler.  I have then added code to the `InstanceCatalogWriter` so that, for each InstanceCatalog that gets written, it will write out a truth catalog with the RA, Dec, SED, magNorm, uniqueId, and galaxy_id of objects for which `is_sprinkled==1`.  These are just placeholder columns.  We will need to talk about what we want the columns to be.

Obviously, this PR will result in many truth catalogs (one per InstanceCatalog) being written.  We will have to add code elsewhere to cull these down into a single, multi-epoch truth catalog, assuming this is the strategy we want to adopt for sprinkler truth info.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -5,6 +5,7 @@ import copy
 from .SedFitter import sed_from_galacticus_mags
 from .SedFitter import sed_filter_names_from_catalog
 from lsst.utils import getPackageDir
+from lsst.sims.catalogs.definitions import InstanceCatalog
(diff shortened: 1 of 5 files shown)


danielsf:
Note: I haven't been able to test this on sprinkled supernovae, because I haven't found an obsHistID that contains a sprinkled supernova.  If someone knows of one of the top of their head, please let me know.

danielsf:
#52 supersedes this PR

PR #51, opened 2018-06-04
add ability to cut on structure function

danielsf:
when creating AGN variability model

Demanding that the structure function is <= 4.0 magnitudes reduces the number of AGN from 328,872 down to 326,505 (in both cases, the cut on black hole mass was 10**7 solar masses and the cut on m_i was 30.0)


jchiang87 (approved):
looks good to me.

PR #52, opened 2018-06-07
U/danielsf/separate/objects

danielsf:
This PR expands #49 so that, when writing PhoSim InstanceCatalogs, disks, bulges, and AGN/SNe all get written to separate gzipped txt files which are passed to PhoSim through `includeobj`.  This is to make life easier on the generic catalog reader when it tries to load InstanceCatalogs.

This PR also supersedes #47.  I was touching a lot of the same lines, so I just tried to fold in Jim's work removing the customized ICRS catalog classes.


danielsf:
PR #49 was held up because I was not sure that it would handle sprinkled SNe correctly.  I have since tested this code and it does handle them as it should.

We will still need the following Twinkles PR

https://github.com/LSSTDESC/Twinkles/pull/472

in order for this PR to work

cwwalter:
Seperating components sounds great. But, I wonder if it is a good idea to seperate bulge and disks. Theses are really the same object and it will make dealing with theinstance catalogs and doing manual checks or manually pulling a few galaxies out of the instance catalogs with standard Unix tools for examination or simulation tedious and error prone.

danielsf:
I can go either way on the "do we separate bulges and disks" question.  I just noticed that, in Javier's InstanceCatalog reader, he included code to inspect the uniqueId to figure out which components were bulges and which were disks.  One argument in favor of separating the components: Javier finds bulges and disks by looking at `uniqueId%1024` and comparing to the offset applied to `uniqueId` for each component to figure out which are bulges and which are disks.  This offset is somewhat arbitrary and will almost certainly change in cosmoDC2 when we stop using the protoDC2 database emulator classes.  Conversely, if the components were separate and you needed to figure out which bulges and disks to compare, you know that you are always looking for objects with identical `uniqueId//1024` (i.e. the 1024 is less likely to change than the offset).

yymao:
It certainly would make the reader a bit easier to write if bulges and disks are in separate files. But the particular part of the code @danielsf mentioned above can be configurable (i.e. we can specify in the config file the corresponding between `uniqueId%1024` and disk/bulge). Hence, either way would still work.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -385,19 +427,18 @@ def make_instcat_header(star_db, obs_md, outfile, object_catalogs=(),
...
-    if imsim_catalog:
-        cat.phoSimHeaderMap['rawSeeing'] = ('rawSeeing', None)
-        cat.phoSimHeaderMap['FWHMgeom'] = ('FWHMgeom', None)
-        cat.phoSimHeaderMap['FWHMeff'] = ('FWHMeff', None)
-    else:
-        cat.phoSimHeaderMap['camconfig'] = 1
+    cat.phoSimHeaderMap['camconfig'] = 1
+
+    # the following commands are needed by imSim
+    cat.phoSimHeaderMap['rawSeeing'] = ('rawSeeing', None)
```
jchiang87: These are not needed by imSim any longer since it will calculate these quantities if they aren't in the instance catalog.

jchiang87 (changes requested):
Looks good to me.  The only change would be to remove the lines writing `rawSeeing`, `FWHMgeom`, and `FWHMeff`.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: done

danielsf:
@jbkalmbach Just want to make sure you saw this, since I issued it while you were at the T/VS working group meeting

jbkalmbach:
Oh, sorry I did miss this. I will take a look in the morning.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -207,13 +206,18 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
         # Ensure that the directory for GLSN spectra is created
         os.makedirs(glsn_spectra_dir, exist_ok=True)
 
-        cat_name = 'phosim_cat_%d.txt' % obsHistID
+        phosim_cat_name = 'phosim_cat_%d.txt' % obsHistID
         star_name = 'star_cat_%d.txt' % obsHistID
         bright_star_name = 'bright_stars_%d.txt' % obsHistID
         gal_name = 'gal_cat_%d.txt' % obsHistID
         knots_name = 'knots_cat_%d.txt' % obsHistID
         #agn_name = 'agn_cat_%d.txt' % obshistid
```
jbkalmbach: Do we need this? Can we remove this?

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -278,35 +277,65 @@ def write_catalog(self, obsHistID, out_dir='.', fov=2):
...
+            written_catalog_names.append(cat_name)
         else:
 
-            self.compoundGalICList = [self.instcats.DESCQACat_Bulge, self.instcats.DESCQACat_Disk,
-                                      self.instcats.DESCQACat_Twinkles]
+            class SprinkledBulgeCat(SubCatalogMixin, self.instcats.DESCQACat_Bulge):
+                subcat_prefix = 'bulge_'
+
+                # must add catalog_type to fool InstanceCatalog registry into
+                # accepting each iteration of these sprinkled classe as
```
jbkalmbach: typo classes

jbkalmbach (approved):
Looks good. Just some clean up comments added.

One question though that has probably been answered in the DC2 repo and I missed it. What is the actual difference between the truth catalogs and the instance catalogs? It seems to me that the truth catalogs are just subsets of information in the instance catalogs. Is this true and am I remembering correctly that this would be because the instance catalogs will go away and we want something smaller to save?

danielsf:
My understanding was that the truth catalogs were meant to be a way to explicitly separate out sprinkled objects from unsprinkled objects.  Truth information for unsprinkled objects can be harvested from the extragalactic catalog after image generation, but, since the sprinkler will replace some cosmoDC2 galaxies with OM10 galaxies, we needed a way to know which objects had been replaced by the sprinkler.  I may be misunderstanding something, though.

danielsf:
I no one objects, I will merge this at noon today.

cwwalter:
So, just to be clear, going forward after this bulges and disks will be in their own object.gz files.  Is that correct?

danielsf:
That is correct.

danielsf:
I wasn't sure where that conversation was, though. It will be easy to put bulges and disks back together if we decide that is better.

danielsf:
@jbkalmbach am I good to merge the supporting PR on the Sprinkler?

jbkalmbach:
Yeah, I don't have any objections. I was just trying to get more information on what each output is going to be used for and maybe that discussion is already/should be somewhere in the DC2 repo.

PR #53, opened 2018-06-20
U/danielsf/shear/convention

danielsf:
This pull request changes the interface with the shear components so that shear_2_phosim gets written to the InstanceCatalogs.

It also fixes a bug in which the user-specified descqa_catalog was not being passed to the DBObject classes used by the sprinkler.


jchiang87 (approved):
Thanks, Scott.  This looks good to me.

danielsf:
Cool.  I will merge at the end of the day

PR #54, opened 2018-07-12
U/danielsf/truth/catalogs

danielsf:
This PR contains the code I used to generate the protoDC2 truth catalogs.  A lot of it (especially the time domain part) is not ready for public consumption (we need to have an intentional discussion about how we are going to provide truth information for time-varying sources).  I would like to merge this, however, so that I have a baseline to work against when I start developing for CosmoDC2 truth catalogs.


danielsf:
@jbkalmbach I would like to be able to pass this code on to other so that they can start thinking about how to generate a truth catalog for cosmoDC2.  Do you mind giving this PR a quick review?  Thanks.

workspace/truth/generate_truth_params.py
```diff
@@ -0,0 +1,82 @@
...
+                                                 agn_db=args.agn,
+                                                 yaml_file=args.yaml,
+                                                 out_dir=args.sql_dir,
+                                                 out_file=args.sql_file)
+
+print('\nwrote\n%s\n' % sql_file_name)
+print('fov %e in %e sec' % (args.fov, time.time()-t_start))
+exit()
+
+### NOTHING BELOW HERE ACTUALLY HAPPENS ###
```
jbkalmbach: Is this for testing? And is it still useful?

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
```diff
@@ -0,0 +1,333 @@
...
+        cmd += '''(uniqueId int,
+                   obshistid int, mag float)'''
+        cursor.execute(cmd)
+
+        cmd = '''CREATE TABLE sne_lc '''
+        cmd += '''(uniqueId int,
+                   obshistid int, mag float)'''
+        cursor.execute(cmd)
+
+        cmd = '''CREATE TABLE obs_metadata '''
```
jbkalmbach, line 149: I don't know if this is too tough to do with sqlite in python, but maybe you could make this adaptive so that if people need extra columns from the opsim database they could add them into this table? When making light curves for Ji Won she needed `fiveSigmaDepth` and I can imagine other people might want different columns if they want to do things with the light curves.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
```diff
@@ -0,0 +1,333 @@
...
+    (htmid_dict,
+     mjd_dict,
+     filter_dict) = get_pointing_htmid(pointing_dir, opsim_db_name,
+                                       ra_colname=ra_colname,
+                                       dec_colname=dec_colname)
+    t_htmid_dict = time.time()-t_start
+
+    with sqlite3.connect(out_file_name) as conn:
+        cursor = conn.cursor()
+        values = ((obs, mjd_dict[obs], filter_dict[obs])
```
jbkalmbach: I had to change `obs` to `int(obs)` here to get the `obsHistId` to appear correctly in the light curve db.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
```diff
@@ -0,0 +1,333 @@
...
+    agn_dtype = np.dtype([('uniqueId', int), ('galaxy_id', int),
+                          ('ra', float), ('dec', float),
+                          ('redshift', float), ('sed', str, 500),
+                          ('magnorm', float), ('varParamStr', str, 500)])
+
+    agn_base_query = 'SELECT uniqueId, galaxy_id, '
+    agn_base_query += 'raJ2000, decJ2000, '
+    agn_base_query += 'redshift, sedFilepath, '
+    agn_base_query += 'magNorm, varParamStr '
+    agn_base_query += 'FROM zpoint WHERE is_agn=1 '
```
jbkalmbach, line 305: I think you need a `agn_base_query += 'AND is_sprinkled=1 '` here.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
```diff
@@ -0,0 +1,333 @@
...
+
+            agn_query = agn_base_query + 'AND htmid=%d' % htmid
+
+            agn_iter = db.get_arbitrary_chunk_iterator(agn_query,
+                                                       dtype=agn_dtype,
+                                                       chunk_size=10000)
+
+            for i_chunk, agn_results in enumerate(agn_iter):
+                values = ((int(agn_results['uniqueId'][i_obj]),
+                           int(agn_results['galaxy_id'][i_obj]),
```
jbkalmbach, line 367: This needs to be `int(np.floor(agn_results['galaxy_id'][i_obj] / 10000))` for now to get the proper `galaxy_id` from a sprinkled object. And we'll have to adjust it when I finalize the number scheme for `cosmoDC2_v1.0`.

workspace/truth/generate_truth_params.py
```diff
@@ -0,0 +1,82 @@
...
+from desc.sims.GCRCatSimInterface import write_sprinkled_lc
+from lsst.sims.utils import ObservationMetaData
+from lsst.sims.photUtils import BandpassDict
+
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--sql_dir', type=str, default=None)
+parser.add_argument('--sql_file', type=str, default='sprinkled_objects.sqlite')
+parser.add_argument('--fov', type=float, default=None)
```
jbkalmbach, line 34: Would be useful to add `field_ra` and `field_dec` params here.

jbkalmbach (changes requested):
Skimmed the parts I haven't really interacted with and they seem fine on a quick look. Mostly focused on the AGN and sprinkled light curve code that I've used and left comments there. There's a couple changes that I think you should include (about the obsHistIds and the sprinkled galaxy_ids) and then some that are more just comments.

workspace/truth/generate_truth_params.py
danielsf: Not exactly testing, but definitely not useful.  I am removing it.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
danielsf: I would like to keep the output from this as lightweight as possible.  If people want other information from the OpSimDB, they should be able to just query that database directly and use `np.in1d` or `np.searchsorted` to effectively join the two tables.  If you think that is too slow or awkward, we can revisit the question in another issue, if you don't mind.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
danielsf: Ah... yes.  `sqlite` hates `np.int`.  Thanks.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
danielsf: Either that, or I need to rename the table.  I may have originally intended this table to just contain sprinkled objects, but now I think it should contain light curves for all AGN and SNe, sprinkled or otherwise.

Do you think there is value in keeping the sprinkled light curves separate from the unsprinkled light curves, or will people be able to discern them based on position relative to the DDF?

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
danielsf: Or should we leave it as is?  If we just write the raw `galaxy_id` to the table here, that would be a way for people to identify sprinkled AGN vs. unsprinkled AGN.  I suppose they would then have to know how to disentangle the true host `galaxy_id` from the sprinkled `galaxy_id`.  Do we think people are going to want to associate the truth_catalog entries with the extragalactic catalog entries?

workspace/truth/generate_truth_params.py
danielsf: Done

danielsf:
If you want to come by and chat about this, feel free.  It was a very stream of consciousness design leading up to the CMU meeting.  It's possible that you are right and I should just implement your comments.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
evevkovacs: I think people might want to match truth catalog entries back to the extragalactic catalog entries, so there should be some simple way to extract the  host galaxy_id from the sprinkled galaxy_id.

python/desc/sims/GCRCatSimInterface/TruthCatalogLC.py
jbkalmbach: What about adding the `is_sprinkled` column and then putting them all in one table? That would make it easy to search within sqlite for just the sprinkled objects.

jbkalmbach: Haha, I just noticed that's what you ended up doing. Maybe I should go through the diff first...

jbkalmbach (approved):
Looks good. Thanks for adding in the sprinkled column.

danielsf:
;)

Thanks for the quick review of the new commits.

You might see a few more come through in the next few hours.  I need to test that the `sprinkled` flag works correctly.  I also want to include the unsprinkled SNe before I hand this off to Eve and Yao.  Don't feel the need to review things as they come through, but if you notice something odd, let me know.

yymao:
@danielsf I encountered an issue when trying to make the light curve reader work (cf https://github.com/LSSTDESC/gcr-catalogs/issues/203). I was assuming that all objects that are present in `variables_and_transients` would also be present in `light_curves`, but apparently that is not the case. 

Here's an example with `/global/cscratch1/sd/danielsf/Run1.2_truth/run_1.2_trial_light_curves_181004.db`:

```sql
sqlite> select uniqueId from variables_and_transients limit 1;
9262
sqlite> select obshistid from light_curves where uniqueId=9262;
sqlite> 
```

Is this expected or is this due to a bug in the truth catalog generation?

danielsf:
@yymao It's not unexpected (which doesn't mean it isn't also a bug)

When I populate the `variables_and_transients` table, I just write out all of the objects in our simulated population.  This neglects two details

1) To make it into the `light_curves` table, an object actually has to appear on a detector.  It is possible that there are supernovae that explode, rise, and fade before our simulated survey ever looks at them (or while it is looking away).

2) The population of simulated SNe we used for protoDC2 covers an area of sky larger than the protoDC2 area, so, even if we caught every supernova in protoDC2, there are going to be supernovae that do not appear in `light_curves` because they are outside of our survey area.

If this population of unrealized light curves is a problem for you, I can re-create the variables catalog without them (I was unaware of (2) when I wrote the code to create the truth catalog; it should be straightforward to omit the objects that run afoul of (1)).

Let me know.

yymao:
I can work around this in the reader now that I know it is expected. However, this means that if an user asks for the light curves of all variable objects, say within an RA range, the user may get some empty arrays for some objects. It's not very initiative but it's ok. 

How feasible is it to add a column in `variables_and_transients` table that contains the number of rows in the `light_curves` table for that object (i.e., the number of detections of that object), so that the user can select on this column?

danielsf:
Adding an `n_detections` column should be pretty easy.  Do you want that for Run 1.2 or should we add it to the list for Run 2.0?

yymao:
Doesn't matter to me actually, in terms of getting the reader to work. The reader code would be essentially the same. It's just that without this column the users won't be able to select only objects that have light curves. 

So I guess the question is whether people need this for their use cases.

danielsf:
If no one objects, I am going to merge this around noon PDT on October 11, just so that we have a starting point from which to start improving the truth catalog infrastructure.

PR #55, opened 2018-07-22
U/danielsf/cosmodc2/healpix

danielsf:
This PR speeds up InstanceCatalog generation so that it is feasible to run on CosmoDC2.  Specifically, speed-ups were achieved by

- applying a native_filter on healpixel ID
- only loading each cosmoDC2 quantity once, and holding all required values in memory
- using scipy.spatial.cKDTree to do the nearest neighbor search in SED assignment

With these optimization, a single LSST field of view (radius=2 degrees) was generated in ~ 70 minutes with a maximum memory footprint on Cori's login node of 20GB.  This test did not use the sprinkler.

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
+++ b/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
@@ -113,18 +113,10 @@ class DESCQAChunkIterator(object):
                              if descqa_catalog.has_quantity(self._column_map[name][0])]
 
             self._loaded_qties = {}
-            n_names = len(qty_name_list)
-            d_name = 10
-            for first_name in range(0, n_names, d_name):
-                last_name = first_name + d_name
-                if n_names-last_name < d_name:
-                    last_name = n_names
-
-                raw_qties = descqa_catalog.get_quantities(qty_name_list[first_name:last_name],
+            for name in qty_name_list:
+                raw_qties = descqa_catalog.get_quantities(name,
                                                           native_filters=native_filters)
-
-                for name in raw_qties.keys():
-                    self._loaded_qties[name] = raw_qties[name][self._data_indices]
+                self._loaded_qties[name] = raw_qties[name][self._data_indices]
 
             print('loaded all quantities')
             # since we are only keeping the objects that will ultimately go into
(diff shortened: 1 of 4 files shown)


python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -157,8 +189,32 @@ def _init_data_indices(self):
             except (TypeError, IndexError):
                 radius_rad = self._obs_metadata._boundLength
 
-            ra = self._descqa_obj._catalog['raJ2000']
-            dec = self._descqa_obj._catalog['decJ2000']
+            if 'healpix_pixels' in dir(descqa_catalog):
```
yymao: Maybe do instead:
```python
if 'healpix_pixel' in descqa_catalog._native_filter_quantities:
```
The fact that `healpix_pixels` is in `descqa_catalog` is kind of implementation detail that may not persist.

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -157,8 +189,32 @@ def _init_data_indices(self):
...
+                ra_rad = self._obs_metadata._pointingRA
+                dec_rad = self._obs_metadata._pointingDec
+                vv = np.array([np.cos(dec_rad)*np.cos(ra_rad),
+                               np.cos(dec_rad)*np.sin(ra_rad),
+                               np.sin(dec_rad)])
+                healpix_list = healpy.query_disc(8, vv, radius_rad,
+                                                 inclusive=True,
+                                                 nest=False)
+
+                print('healpix list ',healpix_list)
```
yymao: How about? Not shorter but slightly safer?
```python
healpix_filter = None
for hh in healpix_list:
    filter_this = GCRQuery('healpix_pixel==%d' % hh)
    if healpix_filter is None:
        healpix_filter = filter_this
    healpix_filter |= filter_this
```

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -167,6 +223,8 @@ def _init_data_indices(self):
         if self._chunk_size is None:
             self._chunk_size = self._data_indices.size
 
+        return native_filters
```
yymao: both `self._data_indices` and `self._chunk_size` now depend on `native_filters`, so all should be returned or stored in `self`.

yymao (changes requested):

danielsf:
I'll rerun the test to make sure I didn't accidentally break anything while responding to the review; I'll ping you when that passes.

yymao (approved):
LGTM

PR #58, opened 2018-09-05
Updates the knots production code to be compatible with recent GCR version

EiffL:
Due to a number of updates of the GCR in recent months, the code for producing the knots didn't work anymore. This PR repairs the pieces that were out of date and sims_GCRCatSimInterface can produce instance catalogs with knots again.

Note that this doesn't touch interactions with the sprinkler, that's a different issue (it doesn't work yet...).

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
+++ b/python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
@@ -5,6 +5,7 @@ __all__ = ["DESCQAObject", "bulgeDESCQAObject", "diskDESCQAObject", "knotsDESCQA
            "deg2rad_double", "arcsec2rad", "SNFileDBObject"]
 
 import numpy as np
+import re
 import healpy
 from lsst.sims.catalogs.db import fileDBObject
 
@@ -375,10 +376,12 @@ class DESCQAObject(object):
 
         # Apply flux correction for the random walk
         add_postfix = []
+        disk_re = re.compile(r'sed_(\d+)_(\d+)_disk_no_host_extinction$')
         for name in gc.list_all_native_quantities():
             # To account for the new composite catalog API,where name is a tuple
             parent, name = name
-            if 'SEDs/diskLuminositiesStellar:SED' in name:
+            disk_match = disk_re.match(name)
+            if disk_match is not None:
                 # The epsilon value is to keep the disk component, so that
                 # the random sequence in extinction parameters is preserved
                 eps = np.finfo(np.float32).eps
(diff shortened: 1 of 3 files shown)


python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -5,8 +5,10 @@
            "deg2rad_double", "arcsec2rad", "SNFileDBObject"]
 
 import numpy as np
+import re
```
yymao: no longer needed

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -352,8 +345,11 @@ def _transform_catalog(self, gc):
         additional_postfix = ()
 
         # Test for random walk specific addon
-        if _ALPHA_Q_ADD_ON_IS_AVAILABLE:
-            if isinstance(gc, AlphaQAddonCatalog):
+        # This gets activated only if the catalog is a composite, and only if one
+        # of the composite catalogs is a knots catalog
+        cat_info = gc.get_catalog_info()
+        if cat_info['subclass_name'] == 'composite.CompositeReader':
+            if 'knots' in [c['catalog_name'] for c in cat_info['catalogs']]:
```
yymao: Lines 351-352: protect edge cases:
```python
    if (cat_info.get('subclass_name') == 'composite.CompositeReader' 
        and any('knots' in c.get('catalog_name', '') for c in cat_info.get('catalogs', []))):
```

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -381,8 +377,12 @@ def _transform_knots(self, gc):
 
         # Apply flux correction for the random walk
         add_postfix = []
+        
         for name in gc.list_all_native_quantities():
```
yymao: Line 381: I think this needs to be changed to:
```python
    for name in gc.list_all_quantities(include_native=True):
```
so that it returns both native and derived quantities.

yymao (changes requested):
@EiffL see my inline comments

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
EiffL: I won't argue with your  superior Python skills :-)

yymao (approved):

danielsf (approved):

EiffL:
Thanks both !

PR #59, opened 2018-09-06
Minor fix that prevents annoying crashes when the  sprinkler is disabled

EiffL:
In the current code, when the sprinkler is disabled, the `generateInstCat.py` will crash if the path to `SIMS_GCRCATSIMINTERFACE_DIR/data/outputs` doesn't exist, which shouldn't be a problem given that the sprinkler is disabled. This PR just fixes the path verification logic and ignores the missing path in case the sprinkler is disabled


danielsf:
I know you didn't ask for my review, but this looks fine to me.  Given the clock that is ticking over Run 2.0i, I say "merge it".

mpwiesner:
Looks good, thanks.

mpwiesner (commented):
Looks good.

EiffL:
Thanks  guys  :-)

PR #60, opened 2018-09-06
Homogenizes the GCR SED extraction between SEDFitter and DatabaseEmulator

EiffL:
In the case of a composite catalog, `gc.list_all_quantities(include_native=True)` returns a list of tuples or simple string, in the latter case the loop over quantities affected by this  PR would fail. So here is a small fix that simply reuses the same code as in SEDFitter.py


EiffL:
This is just to correct a small unintended consequence of a fix to the  last PR

yymao (approved):
LGTM

PR #61, opened 2018-09-11
U/danielsf/lensing/ra/dec

danielsf:
This PR addresses several issues:

- It makes sure we are using the lensed RA and Dec of objects, rather than their true RA, Dec (PhoSim and ImSim have no way of applying lensing corrections to positions)

- It adds code that bypasses the `FieldRotator` if we set `field_ra=field_dec=0` (this seemed the least invasive way to turn off the code; redirecting the `InstanceCatalogWriter` to use `DatabaseEmulator`s that did not use the `FieldRotator` would have involved many more lines of code

- It rewrites the supernova InstanceCatalog generation code to read the supernova parameters from a sqlite database (and provides a script to create that database from csv files that Rahul will provide us) 

Note: because the supernovae now use strings for their IDs, this branch must be run against master of `sims_catUtils`.


python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -517,7 +520,73 @@ class SNFileDBObject(fileDBObject):
                ('Tx0', 'x0_in'),
                ('Tx1', 'x1_in'),
                ('Tc', 'c_in'),
-               ('id', 'snid_in'),
+               ('id', 'snid_in', str, 20),
```
jchiang87: Will this be the object id in the instance catalog?   So for the example files Rahul provided, they would be object ids like `MS_564_7` indicating SNe 7 in the main survey region in healpixel 564?

jchiang87 (approved):
Looks good, though I do have the one question.  Thanks for this, Scott!

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
danielsf: Yes.  I should probably bump the character limit up, just in case.

PR #62, opened 2018-09-13
Disable knots below a given magnitude cuts

EiffL:
@danielsf @jchiang87  I have implemented a quick and dirty fix, I'm running out of time to refine it. Basically, it will not modify the flux of the disks if the galaxy is faint enough, and set 0 flux to the knots. I hope ImSim will be able to simply discard these 0 flux entries. 
@danielsf Do you think this will work ok ? Also, I would need you advice on the best way to pass the magnitude limit to the code, right now it's hardcoded.

This hasn't be tested yet!

### DIFF
--- a/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
+++ b/python/desc/sims/GCRCatSimInterface/CatalogClasses.py
@@ -197,7 +197,9 @@ class PhoSimDESCQA(PhoSimCatalogSersic2D, EBVmixin):
     # below are defined getter methods used to define CatSim value-added columns
     @cached
     def get_hasDisk(self):
-        output = np.where(self.column_by_name('stellar_mass_disk')>0.0, 1.0, None)
+        output = np.where(np.logical_and(self.column_by_name('stellar_mass_disk')>0.0,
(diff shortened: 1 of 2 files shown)


danielsf:
Would it be safer to edit the `get_hasKnots()` method in `CatalogClass.py` to be

```
def get_hasKnots(self):
    has_disk = self.column_by_name('hasDisk').astype(float)
    imag = self.column_by_name('mag_i_lsst')
    return np.where(np.logical_and(np.isfinite(has_disk), imag<=KNOTS_MAG_CUTOFF), has_disk, np.NaN)
```

I'm worried about what the InstanceCatalog will look like if `hasDisk==1` but knot flux is zero.  In theory the `magNorm` for the knots should come back as a NaN, in which case the knots would be omitted, but I'm not sure of that.

danielsf:
Or is it that the disk SED fluxes in `cosmoDC2_v0.1_image_addon_knots` have already been adjusted and we need to undo that?

danielsf:
@jchiang87 This is ready for you to look at.  It works.  Flux is conserved to one part in 10**7 in the case where knots are removed

jchiang87 (approved):
lgtm

PR #64, opened 2018-09-16
Adds instance catalog crawler

EiffL:
@jchiang87 @danielsf  Adds a script that crawl instance catalogs and dropped offensive objects, or apply last minute fixes.
WIP: Not ready to merge yet !
I'm checking that it works, then will improve the code that actually opens and outputs instance catalogs.

### DIFF
--- a/bin.src/instance_crawler.py
+++ b/bin.src/instance_crawler.py
@@ -25,10 +25,10 @@ def fopen(filename, **kwds):
     """
     abspath = os.path.split(os.path.abspath(filename))[0]
     try:
-        if filename.endswith('.gz'):
-            fd = gzip.open(filename, **kwds)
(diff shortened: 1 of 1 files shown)


danielsf (approved):
Looks good to me

EiffL:
@rmandelb @jchiang87 @danielsf Ok, this should be final now, I'm doing some tests but it all looks good. Number of lines are preserved, randomly picked entries look the way I would expect, flux correclty preserved between knots and disks after correction

katrinheitmann:
One quick question: where are you going to write those files to? Same 
directories as the original files are in and name them differently?

Then I would have to copy a lot of extra data since I would want to just 
copy full directory structures.

If the plan is to do that, then I think we should still tar also the 
unchanged files so that we don't have a flood of data.

Thanks!

PR #65, opened 2018-09-20
U/danielsf/generate/master

danielsf:
This PR includes my script to add `phosim_*.txt` to InstanceCatalogs that did not run to completion


danielsf:
I forgot all of the other changes that went into this PR.  I don't think we should try to merge those, yet.

PR #68, opened 2018-09-28
U/danielsf/rebase/config

danielsf:
This PR implements a lot of changes in the InstanceCatalog writing API.

- It alters `InstanceCatalogWriter.py` to output a detailed job log indicating which components (stars, bulges, disks, knots, AGN, SNe) have been completed after how much clock time

- It also makes it possible to pass in parameters to `generateInstCat.py` using a file containing a json-ized dict of prameters.  This makes slurm batch scripts that submit many InstanceCatalog generating jobs with identical parameters much cleaner.  I wasn't terribly careful when I thought through this function.  I just wanted something that would work during the big DC2 crunch weekend.  Please scrutinize my decisions.  There should be an example config file here

https://github.com/LSSTDESC/sims_GCRCatSimInterface/blob/u/danielsf/rebase/config/workspace/parallel/config_file_edison.json

Finally: some changes are made to allow us to use the faster AGN simulation techniques in the `u/danielsf/cosmodc2/agn` PR.


bin.src/generateInstCat.py
```diff
@@ -33,6 +34,22 @@ def generate_instance_catalog(args=None, lock=None):
             warnings.filterwarnings('ignore', 'invalid value', RuntimeWarning)
 
         if not hasattr(generate_instance_catalog, 'instcat_writer'):
+            config_dict = {}
```
jchiang87, line 37: It looks like setting these entries in `config_dict` is intended to document the command line arguments in the `status_file`.  In that case, it would be more compact to do
```
config_dict = {}
config_dict.update(args.__dict__)
```
instead of setting each arg explicitly.  The key names for `proper_motion` and `sprinkler` will have `enable_` prepended, which may be what we want anyway.

bin.src/generateInstCat.py
```diff
@@ -120,8 +150,16 @@ def generate_instance_catalog(args=None, lock=None):
...
     parser.add_argument('--job_log', type=str, default=None,
                         help="file where we will write 'job started/completed' messages")
+    parser.add_argument('--pickup_dir', type=str, default=None,
+                        help='directory to check for aborted job logs')
     args = parser.parse_args()
 
+    if args.config_file is not None:
+        with open(args.config_file, 'r') as in_file:
+            config_dict = json.load(in_file)
+        for kk in config_dict:
```
jchiang87: As above, `args.__dict__.update(config_dict)` should work in place of the `for` loop.

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
```diff
@@ -151,6 +154,8 @@ def __init__(self, opsimdb, descqa_catalog, dither=True,
         plc = ParametrizedLightCurveMixin()
         plc.load_parametrized_light_curves()
 
+        self.config_dict = config_dict
```
jchiang87: Even thought we will probably never pass `config_dict=None`, we should probably still do
```
self.config_dict = config_dict if config_dict is not None else {}
```
since there's a `for` loop over `self.config_dict` below.

jchiang87 (changes requested):
I just have a few suggested changes, otherwise this looks fine.

bin.src/generateInstCat.py
danielsf: Done

bin.src/generateInstCat.py
danielsf: Done

python/desc/sims/GCRCatSimInterface/InstanceCatalogWriter.py
danielsf: Done

jchiang87 (approved):

PR #69, opened 2018-10-19
U/danielsf/star/db

danielsf:
This PR adjusts the InstanceCatalog generation pipeline so that it reads in star data from a local sqlite database.  I also included the script to create that database from fatboy.


jbkalmbach (approved):
Looks good to me

PR #73, opened 2019-01-04
U/danielsf/generate/sne

danielsf:
This script should add all of the code needed to generate the database of supernova parameters for DC2 (minus the scripts that create the csv files of hostless SNe, which @rbiswas4 still has somewhere)


danielsf:
I suspect that we need to add more documentation.  Let me know what you think, Rahul.  I would like this code to be in a publicly-available place so that people can know what we did.

PR #75, opened 2019-08-30
Added Nan's changes in sersic_2d routine in om10_lensing_equations.py…

mpwiesner:
….  Also changed np.log to np.log10 in the two lensing programs.  The purpose of Nan's changes was to make the lensing magnification more physical by adding a softening core in the lensed Sersic profile.


linan7788626 (commented):
The changes look good to me. But I am not sure why the below block shows `all checks have failed`, could anyone instruct me to fix the fails?

jbkalmbach (approved):

PR #76, opened 2019-11-08
Write redshift, not redshift_true, to instance catalogs

JoanneBogart:


JoanneBogart:
The only change was to comment out the line overwriting 'redshift' with 'redshift_true'

danielsf:
...ugh.... sorry about that

python/desc/sims/GCRCatSimInterface/DatabaseEmulator.py
```diff
@@ -493,7 +493,8 @@ def _transform_catalog(self, gc):
         """
         self._transform_object_coords(gc)
 
-        gc.add_quantity_modifier('redshift', gc.get_quantity_modifier('redshift_true'), overwrite=True)
+        #gc.add_quantity_modifier('redshift', gc.get_quantity_modifier('redshift_true'), overwrite=True)
```
jchiang87: We should just delete this line instead of commenting it out.

jchiang87 (changes requested):
Aside from my comment, this looks good to me.  The noMW flux calculations using the instance catalogs made with this change show that the resulting fluxes match to high precision the corresponding truth catalog values.

PR #80, opened 2019-11-15
Assigning filter-dependent tau for AGN variability

jiwoncpark:
Based on the slides I presented on the run 2.2i DDF AGN variability in the SSim telecon on 11/14, it seems that assuming a filter-independent tau (the damping timescale parameter) may be contributing to the problem of tau being small, by a factor of 3, compared to the S82 quasars even after matching the depths. This PR assigns filter-dependent tau in the same manner as what we do for SF_inf.

### DIFF
--- a/bin.src/create_agn_db.py
+++ b/bin.src/create_agn_db.py
@@ -307,7 +307,7 @@ if __name__ == "__main__":
                              tau_dict['y'])
 
             vals = []
-            for gal_id_i, htmid_i, mag_norm_i, seed_arr_i, 
+            for (gal_id_i, htmid_i, mag_norm_i, seed_arr_i, 
                 sf_u, 
                 sf_g, 
                 sf_r, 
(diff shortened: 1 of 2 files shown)


bin.src/create_agn_db.py
```diff
@@ -277,26 +280,64 @@ def get_m_i(abs_mag_i, redshift):
...
+                                                   sf_y, 
+                                                   tau_u, 
+                                                   tau_g, 
+                                                   tau_r, 
+                                                   tau_i, 
+                                                   tau_z, 
+                                                   tau_y)
+                             ))
+
+                cursor.executemany('INSERT INTO agn_params VALUES(?, ?, ?, ?)', vals)
```
jchiang87: This line and the one below is indented one block too far.  It should be outside of the immediately preceding `for` loop.

jiwoncpark: Fixed!

PR #85, opened 2020-02-19
Various fixes and mods to create_agn_db.py

jchiang87:
* Use `redshift` instead of `redshift_true`.
* Modify `tau_from_params` and `SF_from_params` to provide parameter values for a sequence of effective wavelengths using a single `rng` state.
* Select on `duty_cycle_on` for entries written to `agn_params` table.
* Add `redshift`, `M_i`, `ra`, and `dec` columns to `agn_params` table.
* Fix travis-ci builds using docker image.


bin.src/create_agn_db.py
```diff
@@ -143,11 +147,15 @@ def get_m_i(abs_mag_i, redshift):
         qty_names.append('blackHoleAccretionRate')
         filters.append((lambda x: x>0.0, 'blackHoleAccretionRate'))
 
+    if not args.disable_duty_cycle:
+        qty_names.append('duty_cycle_on')
+        filters.append((lambda x: x, 'duty_cycle_on'))
```
jiwoncpark, line 151: The `duty_cycle_on` flag seems to be different from the `duty_cycle` flag in `/global/projecta/projectdirs/lsst/groups/SSim/DC2/cosmoDC2_v1.1.4/agn_model_191006.h5` -- e.g. in the new catalog with duty cycle applied, only 746969 out of 4221027 AGN have `duty_cycle == True`. Has there been a change to the duty cycle?

jiwoncpark (changes requested):
Thank you, this looks good! Please see a question about the duty cycle flag below. Also, wondering if the default `yaml` option in line 94 should now be cosmoDC2?

bin.src/create_agn_db.py
jchiang87: I'm not familiar with the `duty_cycle` flag.  I used `duty_cycle_on` since it's the only quantity in the cosmoDC2 catalog that has `duty` in the name, i.e.,
```
cat = GCRCatalogs.load_catalog('cosmoDC2_v1.1.4_small_combined_agn')
columns = cat.list_all_quantities()
for item in columns:
    if 'duty' in item:
        print(item)
```
yields
```
duty_cycle_on
```
Selecting on that quantity produces an `agn_params` table with 4221027 entries:
```
[cori13] sqlite3 agn_cosmoDC2_v1.1.4.db
SQLite version 3.29.0 2019-07-10 17:32:03
Enter ".help" for usage hints.
sqlite> select count(*) from agn_params;
4221027
sqlite>
```
Is that the correct number of AGNs?

jchiang87: > Also, wondering if the default yaml option in line 94 should now be cosmoDC2?

Probably the best thing to do would be to make the yaml catalog be a required parameter instead of an option.  I'll go ahead and implement that change.

jchiang87: Regarding the `duty_cycle_on` vs `duty_cycle` issue, I note that there are three `.h5` files in the SSim project area that were created within minutes of each other and one that was created several weeks later:
```
[cori07] pwd
/global/projecta/projectdirs/lsst/groups/SSim/DC2/cosmoDC2_v1.1.4
[cori07] ls -lrt agn*h5
-rw-rw----+ 1 danielsf lsst     1051074212 Sep 18 10:05 agn_mean_mags_duty_cycle.h5
-rw-r-----+ 1 danielsf lsst     7726471942 Sep 18 10:08 agn_mean_mags_backup.h5
-rw-r--r--  1 danielsf danielsf 7757503932 Sep 18 10:11 agn_mean_mags.h5
-rw-r--r--  1 danielsf danielsf 4499351830 Oct  7 08:21 agn_model_191006.h5
[cori07] 
```
`agn_mean_mags.h5` is being used by the `cosmoDC2_v1.1.4_image_combined_agn.yaml` file. See https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/cosmoDC2_v1.1.4_image_combined_agn.yaml and https://github.com/LSSTDESC/gcr-catalogs/blob/master/GCRCatalogs/catalog_configs/AGN_mean_mags_v1.1.4.yaml.  So `agn_mean_mags.h5` is the one that's intended to be used with the cosmoDC2 catalog, and that one has a `duty_cycle_on` column. It looks like Scott created `agn_model_191006.h5` later using different random number seeds and so different AGNs were marked with `duty_cycle == True`.   I don't know why this separate file was needed, but based on the numbers of AGNs with the "duty cycle" turned on, `agn_mean_mags.h5` and `agn_model_191006.h5` are statistically consistent.

jiwoncpark: I see. Then 746,969 must have been the number of AGN that happened to be assigned as `duty_cycle == True` in both random seeds. This makes sense! I'll go ahead and remake the validation plots.

jchiang87:
I believe that this is ready to merge, the distributions having been checked, and the requested change was addressed.  Unless I hear otherwise, I will merge this at 9am PT tomorrow.

jiwoncpark:
@jchiang87 Agreed. For the record, the final validation plots are [here](https://docs.google.com/presentation/d/1BTrklLdyjO6T7wjr9QBjrP9iJZ9MPqUH1sG9vUsIUhM/edit?usp=sharing). (I had uploaded it to the Slack AGN channel earlier.)

PR #87, opened 2020-02-20
add options to skip instcat generation for specified object types

jchiang87:
also add code to skip light curve calculations for unneeded bands.


jbkalmbach (approved):
Looks good to me. Only thing is that it looks like it contains commits from a previous pull request that you already merged to master and I'm not sure how that will affect things.

jchiang87:
I needed to test it in the context of that other branch, so I used that one as the base.  As long as there are no conflicts with current master, I don't think there will be a problem.  If there are, they should be easy to fix.

Unless I hear otherwise, I'll merge this at 5pm PT today.

PR #93, opened 2021-02-25
add the as-run config used for Run3.1i instcats

jchiang87:


JoanneBogart:
I didn't see this until it was already closed.  Is there a reason not to include the as-run config in the repo?

jchiang87:
I made this PR before I remembered that two different versions were used--one for the original Run3.0i and a second for the incremental changes in Run3.1i.  So I wanted to make sure I had correct versions in hand before adding them to the repo.  I wasn't sure when I could get around to that, so rather than have this potentially wrong PR sit around, I closed it.

